PR #28, opened 2020-12-18
Bundle with rollup

dustinbyrne:
Rollup now handles bundling, including the ability to bundle CSS files. This will automatically happen when running `npm run watch`, or during compilation via `npm run compile`.

Other minor highlights:
- SCSS files are moved to `/src` to reflect their presence in the bundling process.
- Yarn is removed in favor of NPM.
- `media/appmap.css` is removed as it is now compiled and emitted to `out/extension.css`.
- ESLint configuration (`eslint.json`) is moved to `eslint.js` and is run continuously during `npm run watch`.


rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+    resolve(),
+    scss(),
+    eslint({
+      exclude: 'src/scss/**',
+    }),
+    sucrase({
+      exclude: ["node_modules/**"],
+      transforms: ["typescript"],
+    }),
+    terser(),
```
kgilpin, line 23: Does this make debugging harder? Adding terser to diagrams-js and models-js has made life harder for me in developing the extension. To me, I don't think the compression is really worth it.

kgilpin (approved):

PR #32, opened 2020-12-29
Remove rollup-plugin-terser

dividedmind:
It causes rollup to produce empty sourcemaps for me. I've spent hours trying to find this problem and then to debug it; in the end
I decided it's not worth the trouble. The extension is still in development, the source code is small and public. We can revisit when needed.


kgilpin (approved):

kgilpin:
Agreed

PR #65, opened 2021-01-06
Bundle scenario.js

dustinbyrne:
Previously, we were loading modules directly from node_modules. This means to publish the VS Code extension we'd need to package all our dependencies. By bundling the JS run in the webview, we can cut down to a single file.

Requires https://github.com/applandinc/models-js/pull/19 and https://github.com/applandinc/diagrams-js/pull/59


kgilpin:
Having a smoke test would make this so much easier to approve!

kgilpin:
@dustinbyrne I'm getting an error while compiling on this branch:

```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
@appland/diagrams (imported by media/scenario.js)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
@appland/diagrams (guessing 'diagrams')
```

It does cause the extension to fail when it loads.

dustinbyrne:
Succeeded by #87

PR #88, opened 2021-01-15
updated gifs, now with SQL

ptrdvrk:
Changes: 001.gif and 002.gif in ./doc/media


dustinbyrne:
We should remove these gifs from source control and put them in an S3 bucket

I'll own this

ptrdvrk:
Thank you Dustin. I'll have better screenshots with S3 and a better ruby appmap soon. I should be able to upload them to S3 buckets, too.

PR #90, opened 2021-01-20
Search for source files by relative path

kgilpin:
This handles two edge cases:
* When the project has source files in subdirs of the main dir
* When appmap-ruby reports absolute paths to Gem files


dustinbyrne (approved):
LGTM

PR #107, opened 2021-01-26
Update categories, add keywords

dustinbyrne:


package.json
```diff
@@ -11,14 +11,24 @@
...
 	"main": "./out/extension.js",
+	"categories": [
+		"Visualization",
+		"Debuggers",
+		"Linters",
+		"Other"
+	],
+	"keywords": [
+		"diagram",
+		"dependency",
```
ptrdvrk, line 27: Can we change "dependency" to "dependency map"? Elizabeth feels very strongly about that. I've found an extension that uses two words as well ("Dependency scan")

ptrdvrk (changes requested):
Please see the comment about the "dependency map" keyword.

ptrdvrk (approved):
Looking good, thank you!

PR #129, opened 2021-02-12
Update README.md

ptrdvrk:
self contained -> standalone extension
added the Troubleshooting section
added "Click on an edge" in the summary of features

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,8 +121,7 @@ Here are some tips:
 
 ### Share AppMap files for viewing in VSCode
 
-Generated AppMap files can be viewed by others with the AppMap extension. So, one option for sharing is to simply send the `appmap.json` file to your colleague. Another option is use of a git repository for sharing the `appmap.json` files with the members of your team.
-
+Generated AppMap files can be viewed by others with the AppMap extension. So, one option for sharing is to simply send the `appmap.json` file to your colleague.
 
 ### Share AppMap files in the SaaS [https://app.land](https://app.land) sandbox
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,7 +121,7 @@ Here are some tips:
 
 ### Share AppMap files for viewing in VSCode
 
-Generated AppMap files can be viewed by others with the AppMap extension. So, one option for sharing is to simply send the `appmap.json` file to your colleague.
+Generated AppMap files can be viewed by others with the AppMap extension. So, one option for sharing is to simply send the `appmap.json` file to your colleague. Another option is use of a git repository for sharing the `appmap.json` files with the members of your team.
```
kgilpin, line 124: This doesn’t seem like a good idea. We don’t really want people putting these big files in Git. How about Drive, Slack or Dropbox instead?

kgilpin (changes requested):

README.md
ptrdvrk: Ok, let me revert this change, the file sharing obviously needs a little more thinking and work.

kgilpin (approved):

dustinbyrne:
In the future please run `npm install` (just the one time is fine) before committing. It'll enable a git post-commit hook that validates your commit message as a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/). This is necessary for automated release. Otherwise it'll wait for a subsequent releasable change to push it out.

ptrdvrk:
Will do Dustin. I am assuming `npm install` is required prior to the final commit before the PR is created?

kgilpin:
Before any commit is made.

PR #164, opened 2021-03-26
fix: Show AppMap instructions on extension first open

sashahonchar:
Fixes: https://github.com/applandinc/appmap-js/issues/136

### DIFF
--- a/web/src/app.js
+++ b/web/src/app.js
@@ -96,15 +96,16 @@ window.addEventListener('error', (event) => {
 window.addEventListener('message', (event) => {
   const message = event.data;
 
-  /* eslint-disable no-case-declarations */
   switch (message.type) {
     case 'update':
-      const { text } = message;
-      app.loadData(text);
+      {
+        const { text } = message;
+        app.loadData(text);
 
-      // Then persist state information.
-      // This state is returned in the call to `vscode.getState` below when a webview is reloaded.
-      vscode.setState({ text });
+        // Then persist state information.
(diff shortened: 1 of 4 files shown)


web/src/app.js
```diff
@@ -22,14 +25,24 @@ app.$on('viewSource', (location) => vscode.postMessage({ command: 'viewSource',
...
-    vscode.setState({ text });
+  /* eslint-disable no-case-declarations */
+  switch (message.type) {
+    case 'update':
+      const { text } = message;
+      app.loadData(text);
+
+      // Then persist state information.
+      // This state is returned in the call to `vscode.getState` below when a webview is reloaded.
+      vscode.setState({ text });
```
dustinbyrne, line 37: If you declare this in its own scope, you won't need to disable the `no-case-declarations` eslint rule. The worry is that a variable is initialized in one case, but then used in a following case and thus never initialized. Adding an explicit scope prevents this:

```js
case 'update':
  {
    const { text } = message;
    app.loadData(text);

    // Then persist state information.
    // This state is returned in the call to `vscode.getState` below when a webview is reloaded.
    vscode.setState({ text });
  }
  break;
```

dustinbyrne (changes requested):

dustinbyrne:
This will also require `@appland/appmap` is upgraded to `v1.10.1`. We've been bumping versions as `fix`. Please let me know if this introduces any issues, we did update some of the bundling between releases.

typaulhus (approved):

PR #180, opened 2021-05-04
feat: send telemetry event after extension installation

sashahonchar:
Fixes: https://github.com/applandinc/appmap-js/issues/200


dustinbyrne:
Rebased into and merged with #188

PR #182, opened 2021-05-08
feat: upload AppMap to AppLand server

artemvaskoo:
**First version of AppMap Uploader**

For test purpose, I had to change `appmap` dependency to be loaded from a local file, as there's no released version yet.

When running `npm install`, remove `node_modules/@appland/appmap` if you encounter installation errors, as it apparently cannot be replaced by a symlink automatically.

After uploading, it displays a notification and opens a browser window, where the uploaded data is collected and added to the user profile. For that to work, you have to use `feature/vscode-upload-handling` branch for AppLand from `Appmap-server` PR#863(https://github.com/applandinc/appmap-server/pull/863) and change `appMap.appLandURL` configuration value to point to your local instance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,11 @@
           "type": "string",
           "default": "",
           "description": "Path for remote recordings"
+        },
+        "appMap.applandUrl": {
+          "type": "string",
+          "default": "https://app.land",
+          "description": "URL of an AppLand cloud instance for AppMap storage"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -202,9 +202,9 @@
     "semantic-release": "semantic-release"
   },
   "devDependencies": {
-    "@babel/core": "^7.12.10",
-    "@babel/preset-env": "^7.12.11",
-    "@babel/preset-typescript": "^7.12.7",
+    "@babel/core": "^7.14.3",
+    "@babel/preset-env": "^7.14.2",
+    "@babel/preset-typescript": "^7.13.0",
     "@semantic-release/changelog": "^5.0.1",
     "@semantic-release/exec": "^5.0.0",
     "@semantic-release/git": "^9.0.0",
@@ -237,6 +237,7 @@
     "style-loader": "^2.0.0",
     "tape": "^5.1.1",
     "temp": "^0.9.4",
+    "terser-webpack-plugin": "^5.1.2",
     "ts-loader": "^8.0.14",
     "ts-node": "^10.0.0",
     "tslib": "^2.1.0",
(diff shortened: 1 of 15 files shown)


src/appmapUploader.ts
```diff
@@ -0,0 +1,45 @@
...
+    const upload_url = this.getUploadUrl();
+
+    axios
+      .post(
+        upload_url,
+        {
+          data: appMapFile.getText(),
+        },
+        {
+          headers: { 'X-Requested-With': 'VSCodeUploader' },
```
kgilpin, line 15: I suggest instead to post the data to an /api route. That will make sense for this flow, and to get a JSON response. It will also be more clear to document.

artemvaskoo: Okay I will change

sashahonchar (approved):

package.json
```diff
@@ -178,7 +188,8 @@
     "webpack-cli": "^4.3.1"
   },
   "dependencies": {
-    "@appland/appmap": "2.3.4",
+    "@appland/appmap": "file:../appmap-js",
```
dustinbyrne, line 191: You can use `npm link` to avoid this
https://docs.npmjs.com/cli/v7/commands/npm-link

package.json
```diff
@@ -178,7 +188,8 @@
...
+    "axios": "^0.21.1",
```
dustinbyrne, line 192: This is a bit nit-picky, but we've included [`bent`](https://www.npmjs.com/package/bent) for making HTTP requests. Do we need this dependency?

package.json
```diff
@@ -75,6 +75,16 @@
         "icon": "$(extensions-filter)"
       }
     ],
+    "configuration": {
+      "title": "AppMap",
+      "properties": {
+        "appMap.appLandURL": {
```
dustinbyrne, line 81: Can we rename this key to `appmap.serverURL`?

dustinbyrne (changes requested):

artemvaskoo:
Updated with telemetry and confirmation

Currently, the confirmation button works globally and will allow any `appmap` to be uploaded after the first one is confirmed. It will need more time to develop if we want to add per-file confirmation.

kgilpin:
Once seems fine to me

kgilpin:
Or simply upload to default URL https://app.land

sashahonchar:
I don't think it's a good idea to hard-code server URL. @kgilpin 
It's better to show an error notification when URL settings is empty. https://github.com/applandinc/vscode-appland/pull/182/commits/15e5cd5adb199af3e5f260a7714cc808e7d69666

kgilpin:
Why would we require the user to enter a setting which is the same for 99%
of users?

sashahonchar:
I agree, when user setting is empty it's better to use https://app.land as default server. 
https://github.com/applandinc/vscode-appland/pull/182/commits/3cbff63bd5b39867bddbe64f71261836f072f457

artemvaskoo:
> Why would we require the user to enter a setting which is the same for 99% of users?
It should not be empty unless it's manually set to empty, because there's a default value for it in `package.json`

dustinbyrne (approved):

Issue #177, opened 2021-05-04
Remote recording actions exist as commands


dustinbyrne:
The command takes a single argument: the target endpoint.
ex: `localhost:3000`, `localhost:8080/myapp`

Commands added:
- [ ] Start recording
- [ ] Stop recording

PR #183, opened 2021-05-12
feat: remote recording commands

sashahonchar:
Fixes: #177 
Fixes: https://github.com/applandinc/appmap-js/issues/201


src/remoteRecording.ts
```diff
@@ -0,0 +1,54 @@
...
+    const request = bent('http://' + baseURL, 'POST', 'json', 200);
+
+    try {
+      await request(this.RECORDING_URI);
+      vscode.window.showInformationMessage(`Recording started at "${baseURL}"`);
+    } catch (e) {
+      vscode.window.showErrorMessage(`Start recording failed: ${e.name}: ${e.message}`);
+    }
+  }
+  static async getStatus(): Promise<void> {
```
dustinbyrne, line 22: Can we split up the HTTP client from the VSCode logic to keep things flexible?

I'd expect the HTTP client to be fit into or be similar to one of the following interfaces:
```ts
interface RecordingClient {
  url: string;

  start(): Promise<void>;
  stop(): Promise<string>;
  status(): Promise<Record<string, unknown>>;
}
```

e.g.:
```ts
const client = new MyRecordingClient('https://localhost:3001');
await client.start();

// ...

const appmap = buildAppMap(await client.stop()).build();
```

Alternatively, as static methods as you currently have it implemented. This probably makes the most sense for this use case. Bonus points for [maintaining functional style](https://www.gamasutra.com/view/news/169296/Indepth_Functional_programming_in_C.php).
```ts
interface RecordingClient {
  start(url: string): Promise<void>;
  stop(url: string): Promise<string>;
  status(url: string): Promise<Record<string, unknown>>;
}
```

The rest of the VSCode logic (`showInputBox`, `showInformationMessage`, etc.) can exist in the command definition.

src/remoteRecording.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+  static async stop(): Promise<void> {
+    const baseURL = await this.getBaseUrl();
+    const request = bent('http://' + baseURL, 'DELETE', 'json', 200);
+
+    try {
+      const response = (await request(this.RECORDING_URI)) as string;
+
+      const document = await vscode.workspace.openTextDocument({
```
dustinbyrne, line 43: I'd like to immediately pop open the AppMap viewer. Any ideas?

An easy fix would be to save the file to disk. Alternatively, a [`TextDocumentContentProvider`](https://code.visualstudio.com/api/extension-guides/virtual-documents#textdocumentcontentprovider) could also work, if the URI ends with `.appmap.json`. Stopping the recording would instead open a file using a custom scheme (something like `appmap-remote://https%3A%2F%2Flocalhost:3000.appmap.json`) which parses the URI, issues a stop request to the endpoint, and returns the JSON document as a `string`.

I'm sure I've left out some details, so please reach out with questions if you'd like some help with this.

dustinbyrne (changes requested):
Looking good! I left a few comments.

src/remoteRecording.ts
sashahonchar: It's not possible to create an object definition like this:
```javascript
const client = new MyRecordingClient('https://localhost:3001');
await client.start();
// ...
const appmap = buildAppMap(await client.stop()).build();
```
The reason is we will ask user for server URL before every command execution.

src/remoteRecording.ts
```diff
@@ -0,0 +1,54 @@
...
+      };
+      const recordingStatus = response.enabled ? 'enabled' : 'disabled';
+      vscode.window.showInformationMessage(`Recording status at "${baseURL}": ${recordingStatus}`);
+    } catch (e) {
+      vscode.window.showErrorMessage(`Recording status failed: ${e.name}: ${e.message}`);
+    }
+  }
+  static async stop(): Promise<void> {
+    const baseURL = await this.getBaseUrl();
+    const request = bent('http://' + baseURL, 'DELETE', 'json', 200);
```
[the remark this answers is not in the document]
sashahonchar: Done https://github.com/applandinc/vscode-appland/pull/183/commits/d81689abff5c3e47f4c9ca9fe2ba07a932e532ae

src/remoteRecording.ts
```diff
@@ -0,0 +1,54 @@
...
+export default class RemoteRecording {
+  private static readonly RECORDING_URI = '/_appmap/record';
+  private static async getBaseUrl(): Promise<string | undefined> {
+    return await vscode.window.showInputBox({
+      placeHolder: 'URL of remote recording server, eg "localhost:3000"'
+    });
+  }
+  static async start(): Promise<void> {
+    const baseURL = await this.getBaseUrl();
+    const request = bent('http://' + baseURL, 'POST', 'json', 200);
```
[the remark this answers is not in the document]
sashahonchar: Done https://github.com/applandinc/vscode-appland/pull/183/commits/6b2744f7e854816acb4d5e77ef938bb36d38946c

src/remoteRecording.ts
dustinbyrne: Sorry, the example I gave is confusing. I don't expect to start/stop in the same block. I'm advocating for separation of the HTTP client and VS code interaction logic:

```ts
context.subscriptions.push(
  vscode.commands.registerCommand('appmap.startRemoteRecording', () => {
    const baseURL = (await getBaseUrl()) || '';
    if (baseURL === '') {
      return;
    }

    try {
      RemoteRecording.start(baseURL);
      vscode.window.showInformationMessage(`Recording started at "${baseURL}"`);
    } catch (e) {
      vscode.window.showErrorMessage(`Start recording failed: ${e.name}: ${e.message}`);
    }
  })
);
```

dustinbyrne (changes requested):
Hi @sashahonchar, I have a few more items that came up earlier today.

- Upon stopping the recording, prompt the user to name the AppMap. This should set the `metadata.name` property before saving the file out or displaying it in the editor.
- A status bar indicator should be present while a recording is in progress. Clicking the indicator stops the recording and prompts the user to enter a name. This action should be able to be cancelled, and the recording continued.
- A history of service URLs saved per-workspace. As a user, I don't want to have to re-enter my service URL every time I start a recording. The user can either type in a service URL or select a previous service URL from a list. The history is unique to the current workspace path.
- A button is added to the AppMap panel to start a new recording. Other button definitions are found here: https://github.com/applandinc/vscode-appland/blob/master/package.json#L116-L125

kgilpin:
When the new AppMap is saved, it should also be processed with the Indexer,
so that it can be efficiently searched and compared.

kgilpin:
I guess you can shell out to the index CLI command to do this.

src/remoteRecording.ts
sashahonchar: Done https://github.com/applandinc/vscode-appland/pull/183/commits/2bdf340efbf2010166c5245dcb69a059c819f61a

src/extension.ts
```diff
@@ -51,4 +60,108 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+
+      try {
+        RemoteRecording.start(remoteURL);
+
+        statusBarItem.text = `$(record) Remote recording is running on ${remoteURL}`;
+        statusBarItem.show();
+
+        vscode.window.showInformationMessage(`Recording started at "${remoteURL}"`);
+      } catch (e) {
+        vscode.window.showErrorMessage(`Start recording failed: ${e.name}: ${e.message}`);
```
dustinbyrne, line 90: This should also clear state. Otherwise, we cannot try again.
```js
remoteURL = null;
```

src/extension.ts
```diff
@@ -51,4 +60,108 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+      }
+
+      remoteURL = (await getBaseUrl()) || '';
+
+      if (remoteURL === '') {
+        return;
+      }
+
+      try {
+        RemoteRecording.start(remoteURL);
```
dustinbyrne, line 83: This call requires `await`, otherwise an exception will never be handled in this scope.

package.json
```diff
@@ -117,11 +130,16 @@
...
         },
         {
           "command": "appmap.findByName",
           "when": "view == appmap.views.local && appmap.hasData",
+          "group": "navigation@1"
+        },
+        {
+          "command": "appmap.startRemoteRecording",
+          "when": "view == appmap.views.local && appmap.hasData",
           "group": "navigation@0"
```
dustinbyrne, line 143: Can we add a `when` condition to display a stop command if a recording is already in progress?

dustinbyrne:
I found a method for displaying an input box with an optional pre-populated item selection. 

```js
import * as vscode from 'vscode';

export default class QuickPickHistory {
  public static async open(): Promise<string> {
    const quickPick = vscode.window.createQuickPick();

    // Add historical endpoints here
    quickPick.items = [{ label: 'a' }, { label: 'b' }, { label: 'c' }];

    // Configure the rest of the input box (placeholder, etc.) 

    return new Promise((resolve) => {
      quickPick.show();
      quickPick.onDidAccept((e) => {
        // If no label is selected, fall back to the input value instead
        console.log(quickPick.selectedItems[0]?.label, quickPick.value);
        resolve(quickPick.selectedItems[0]?.label || quickPick.value);
      });
    });
  }
}
```

sashahonchar:
Fixes: #201

dustinbyrne (approved):

PR #232, opened 2021-06-09
fix: get/set Appmap state

sashahonchar:
This functionality was added in #169 , but later it was removed because there was a bug with not working `setState` command when few Appmaps were opened.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -92,6 +92,33 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
       })
     );
 
+    context.subscriptions.push(
+      vscode.window.registerUriHandler({
+        handleUri(uri: vscode.Uri) {
+          if (uri.path === '/open') {
+            const queryParams = new URLSearchParams(uri.query);
+
+            if (queryParams.get('uri')) {
+              vscode.commands.executeCommand(
+                'vscode.open',
+                vscode.Uri.parse(queryParams.get('uri') as string)
+              );
+            }
+
+            if (queryParams.get('state')) {
+              // let's give some time to open AppMap file before setting state
+              setTimeout(() => {
+                vscode.commands.executeCommand(
+                  'appmap.setAppmapStateNoPrompt',
+                  queryParams.get('state')
+                );
+              }, 1000);
+            }
+          }
+        },
+      })
+    );
+
     registerUtilityCommands(context, properties);
 
     vscode.env.onDidChangeTelemetryEnabled((enabled: boolean) => {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -106,13 +106,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
             }
 
             if (queryParams.get('state')) {
-              // let's give some time to open AppMap file before setting state
-              setTimeout(() => {
-                vscode.commands.executeCommand(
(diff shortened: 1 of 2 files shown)


kgilpin:
Will it be easy to go to a specific event in the Trace view? If I know the
event ID?

sashahonchar:
I think we should create new command for that @kgilpin

dustinbyrne:
Don't we support that already through the `currentView` key?

src/scenarioViewer.ts
```diff
@@ -32,6 +33,8 @@ export class ScenarioProvider implements vscode.CustomTextEditorProvider {
     webviewPanel: vscode.WebviewPanel
     /* _token: vscode.CancellationToken */
   ): Promise<void> {
+    ScenarioProvider.currentWebView = webviewPanel;
```
dustinbyrne, line 36: Is this also executed when switching tabs?

sashahonchar: No, It's not executed when you switch between 2 opened appmap viewers in VSCode

dustinbyrne: Is there a way to listen for that event? Or can we avoid the issue by accepting two arguments instead:
- a path to an AppMap file to be opened by the command
- a state string to be immediately loaded upon opening

sashahonchar: @dustinbyrne , I reviewed VSCode docs and found a solution. https://github.com/applandinc/vscode-appland/pull/232/commits/427e42761d12cb2e556a341ffe1bb5100e31d439

kgilpin:
@sashahonchar is this branch still alive?

I need a command to open the trace view to a specified event.

sashahonchar:
> @sashahonchar is this branch still alive?
> 
> I need a command to open the trace view to a specified event.

@kgilpin , @dustinbyrne  I think this branch should be closed and new issue with detailed description should be opened.

sashahonchar:
@kgilpin, Yeah, I just rebased this branch from master.

dustinbyrne (approved):
I'm happy to merge this as-is, but I think to better support Kevin's use-case, the command should take both a file and state string. Then, once executed, will open the file with the given state. I'll open an issue.

dustinbyrne:
https://github.com/applandinc/vscode-appland/issues/336

src/extension.ts
```diff
@@ -86,6 +86,33 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+            }
+
+            if (queryParams.get('state')) {
+              // let's give some time to open AppMap file before setting state
+              setTimeout(() => {
+                vscode.commands.executeCommand(
+                  'appmap.setAppmapStateNoPrompt',
+                  queryParams.get('state')
+                );
+              }, 1000);
```
dustinbyrne, line 109: This `setTimeout` seems fragile. Is it possible to send this state as part of the command to open the file?

sashahonchar: Good point!
Now we gonna store it in `globalState` and apply after webview load. https://github.com/applandinc/vscode-appland/pull/232/commits/b618f57285e43d50cffd5a25a4f9bcaa539aba0e

dustinbyrne:
I've looked at the code, but I still need to try running this myself. I'll have a chance tomorrow.

PR #233, opened 2021-06-10
feat: Telemetry `install` event

sashahonchar:


src/extension.ts
```diff
@@ -18,6 +20,11 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
   localAppMaps.initialize();
   const { localTree } = registerTrees(localAppMaps);
 
+  if (!context.globalState.get(storeTelemetryInstallKey)) {
+    Telemetry.reportAction('appland.appmap/plugin:install', undefined);
```
dustinbyrne, line 24: `appland.appmap/` is prepended by the library. The first argument only needs to be `plugin:install`.

dustinbyrne (changes requested):

src/extension.ts
sashahonchar: Done https://github.com/applandinc/vscode-appland/pull/233/commits/d68ea06d9ff367a98352b22ce800e494eb690326

dustinbyrne (approved):

PR #234, opened 2021-06-10
Properties: `appmap.project.*`

dustinbyrne:


src/telemetry/properties/versionControlGit.ts
```diff
@@ -0,0 +1,46 @@
+import * as vscode from 'vscode';
+import { PathLike, promises as fs } from 'fs';
+import { join, dirname } from 'path';
+import * as globToRegExp from 'glob-to-regexp';
+import VersionControlProperties from './versionControl';
+
+async function getIgnoreGlobs() {
```
artemvaskoo, line 7: @dustinbyrne , Is there any way to access vscode source control instead of trying to implement custom code for each vcs?

dustinbyrne: Unfortunately vscode doesn't expose a lot of their Git functionality. But, with that being said, we really only need to support `git`. I'm just leaving room for us to expand later if need be.

dustinbyrne:
I've renamed this branch and opened another pull request.
https://github.com/applandinc/vscode-appland/pull/240

PR #242, opened 2021-06-14
`appmap.file` properties

artemvaskoo:
TODO:
- How do we determine `project_config` file type?
- Where do we actually need those?
- Add test


src/telemetry/properties/file.ts
```diff
@@ -0,0 +1,56 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import { workspace } from 'vscode';
+import { createHash } from 'crypto';
+
+const appMapConfigFile = 'appmap.yml';
+const appMapFileExt = '.appmap.json';
```
dustinbyrne, line 7: Prefer `SCREAMING_SNAKE_CASE` for static constants

src/telemetry/properties/file.ts
```diff
@@ -0,0 +1,56 @@
...
+
+function getFileType(baseName: string, dirPath: string): string {
+  if (baseName == appMapConfigFile && dirPath == '.') {
+    return 'appmap_config';
+  } else if (baseName.endsWith(appMapFileExt)) {
+    return 'appmap';
+  }
+  return 'source_code';
+
+  // TODO: How do we determine project_config?
```
dustinbyrne, line 30: I'm working on project properties now, perhaps this should be part of the interface.

PR #249, opened 2021-06-15
feat: Properties: `appmap.version_control.*`

artemvaskoo:
Dives into the private API for a bit, but gets all requested fields without re-implementing any of the git functions or using third party implementations, and that part of API is unlikely to be changed soon


dustinbyrne (changes requested):
I was a bit worried about the usage of a private API, however, according to their [README](https://github.com/microsoft/vscode/blob/main/extensions/git/README.md), "the Git extension exposes an API, reachable by any other extension". So it would seem this is an intended use case, and more of a public API than a private one.

As such, this looks like an acceptable change. It will require unit tests however. Do you mind tackling that before we merge?

artemvaskoo:
I was a bit worried about the usage of a private API, however, according to their [README](https://github.com/microsoft/vscode/blob/main/extensions/git/README.md), "the Git extension exposes an API, reachable by any other extension". So it would seem this is an intended use case, and more of a public API than a private one.

As such, this looks like an acceptable change. It will require unit tests however. Do you mind tackling that before we merge?

src/telemetry/properties/version_control.ts
```diff
@@ -0,0 +1,84 @@
...
+  } catch {
+    return false;
+  }
+}
+
+async function getIgnoredStatus(repo: Repository, filePath: string): Promise<boolean> {
+  try {
+    // Diving into the private API for a bit
+    const ignoreList = (await repo['_repository'].checkIgnore([filePath])) as Set<string>;
+    return ignoreList.has(filePath);
```
dustinbyrne, line 57: OK - this is actually private. We can use this instead: https://github.com/applandinc/vscode-appland/blob/release/appmap-milestones/src/telemetry/properties/versionControlGit.ts#L74-L80

PR #276, opened 2021-07-15
Quickstart auto-open after init

ptrdvrk:
This update is meant for the UX testing exercise. We should have similar functionality for GA, but only open the Quickstart the first time.

### DIFF
--- a/src/tree/milestoneTreeDataProvider.ts
+++ b/src/tree/milestoneTreeDataProvider.ts
@@ -56,6 +56,10 @@ export class MilestoneTreeDataProvider implements vscode.TreeDataProvider<vscode
     return element;
   }
 
+  public getParent(): Thenable<vscode.TreeItem | null> {
+    return Promise.resolve(null);
+  }
+
   public getChildren(): Thenable<vscode.TreeItem[]> {
     const project = this.projects[0];
     if (!project) {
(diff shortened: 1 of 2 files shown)


src/tree/milestoneTreeDataProvider.ts
```diff
@@ -56,6 +56,10 @@ export class MilestoneTreeDataProvider implements vscode.TreeDataProvider<vscode
     return element;
   }
 
+  public getParent(): Thenable<vscode.TreeItem | null> {
```
apotterri, line 59: Is this necessary? It doesn't seem like it's doing much....

ptrdvrk: Without this the milestoneTree.reveal() in extension.ts does not work, errors on missing function getParent().

apotterri: Do you understand why that is? Why wasn't it necessary before your change?

apotterri (changes requested):
The two commits in this PR should get squashed together, which could be done when the PR gets merged. However, they have different semantic release prefixes. Which is correct?

ptrdvrk (commented):
> The two commits in this PR should get squashed together, which could be done when the PR gets merged. However, they have different semantic release prefixes. Which is correct?

The second commit with the "misc:" prefix just updated the comments. It can be removed.

apotterri (approved):

PR #294, opened 2021-07-21
feat: user terminal

sashahonchar:


src/terminal/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.watcher = chokidar.watch(this.baseDir + 'output', {
+      awaitWriteFinish: {
+        stabilityThreshold: 500,
+        pollInterval: 100,
+      },
+    });
+  }
+
+  private exec(command: string): Promise<TerminalResult> {
+    const echoStatus = 'echo $? > ' + this.baseDir + 'exitStatus';
```
kgilpin, line 68: How about using `fs.join` here?

src/agent/appMapAgentRuby.ts
```diff
@@ -68,29 +74,25 @@ export default class AppMapAgentRuby implements AppMapAgent {
...
-          output: true,
-        });
+        const { stderr, exitCode } = await execCommand(path, 'bundle', ['install']);
 
         if (exitCode !== 0) {
           error = stderr;
         }
       }
     } catch (e) {
+      console.log(e);
```
kgilpin, line 95: Can/should this be directed to a VSCode output?

kgilpin:
@sashahonchar you'll probably want to clean up the history before merging.

src/agent/appMapAgentRuby.ts
sashahonchar: Error message will be displayed on Quickstart webview:
https://github.com/applandinc/vscode-appland/blob/7868c2069be47775bbeeb3bdf7dc7c62eeba3b54/src/agent/appMapAgentRuby.ts#L96

src/terminal/index.ts
sashahonchar: Done https://github.com/applandinc/vscode-appland/pull/294/commits/6eb052092ee5a995529d3f1fff62ca54ae0df846

sashahonchar:
> @sashahonchar you'll probably want to clean up the history before merging.

Done. I squashed 'wip' commits into this one https://github.com/applandinc/vscode-appland/pull/294/commits/fd7d8ec02c444498a2b98ab162d5b35b52bf94af

kgilpin (changes requested):
I think it's good that we did this exercise but we need to hold off on replacing the existing flow with this flow for all situations. 

Also, chokidar should be removed and replaced with the built in VSCode file watcher. Native code is not allowed in VSCode extensions.

torchello:
Closing this for now.

PR #307, opened 2021-08-03
Static construction of telemetry events

dustinbyrne:
# wip

- Event definitions move from `src/telemetry.ts` to `src/telemetry/definitions/events.ts`
- Property definitions move from `src/telemetry/index.ts` to `src/telemetry/definitions/properties.ts`
- Metric definitions move from `src/telemetry/index.ts` to `src/telemetry/definitions/metrics.ts`
- Removed `TelemetryContext`, `EventContext` in favor of a `data` argument passed to `TelemetryDataResolver#getValue`. The `data` argument extends `Record<string, unknown>`, and should cover all cases.
- Added `argCache` property to `TelemetryDataResolver`: this property specifies which arguments will be wrapped by a method caching proxy. This means expensive calls can have their return values cached within the same telemetry event.
- `Telemetry.sendEvent` now uses deep generic types and will require all data used by child properties and metrics be supplied.
- `TelemetryDataResolver` can now return a `Record`, which will have it's keys flattened before being appended to the telemetry event data. For example, given the following `Record<string, string>` returned by a data resolver named `appmap.recorder`:
  ```js
  {
    name: 'appmap-ruby',
    version: '0.50.0',
  }
  ```
  The resulting telemetry event data would include:
  ```yml
  appmap.recorder.name = 'appmap-ruby'
  appmap.recorder.version = '0.50.0'
  ```
- The property and metric export strategy has changed. Individual exports exist for each individual `TelemetryDataResolver`, instead of exporting an entire object.


src/scenarioViewer.ts
```diff
@@ -85,6 +85,7 @@ export class ScenarioProvider implements vscode.CustomTextEditorProvider {
             rootDirectory: workspaceFolderForDocument(document)?.uri.fsPath,
             uri: document.uri,
             metadata: JSON.parse(document.getText()).metadata,
+            metrics: message.metrics,
           });
```
dustinbyrne, line 89: Here's what the improved API looks like. The first argument is the `Event` declaration, followed by a union of all the data required by metric and property `TelemetryDataResolvers`.

src/telemetry/definitions/events.ts
```diff
@@ -37,15 +39,16 @@ export const PROJECT_CONFIG_WRITE = new Event({
...
-    Properties.File.SHA_256,
-    Properties.File.METADATA,
-    Properties.Project.LANGUAGE,
+    Properties.FILE_PATH,
+    Properties.FILE_SHA_256,
+    Properties.FILE_METADATA,
+    Properties.PROJECT_LANGUAGE,
   ],
+  metrics: [Metrics.APPMAP_JSON],
 });
```
dustinbyrne: Example `Event` declaration.

src/telemetry/definitions/properties.ts
```diff
@@ -1,144 +1,130 @@
...
+  id: 'appmap.file.path',
+  async value({ uri, rootDirectory }: { uri: Uri; rootDirectory?: PathLike }) {
+    if (!rootDirectory) {
+      // If we cannot resolve a relative path, bail out.
+      return undefined;
+    }
+
+    return path.relative(rootDirectory as string, uri.fsPath);
+  },
+});
```
dustinbyrne, line 27: Example property declaration. Note that some data properties can be optional (`rootDirectory?`), which allows the caller to provide the data if it's available. Circumstances will arise where not all data can be supplied in every context where we want to send a telemetry event. If every field is optional (see `PROJECT_LANGUAGE`), the property or metric will only be attached to the telemetry event if the data is provided.

src/telemetry/definitions/properties.ts
```diff
@@ -1,144 +1,130 @@
...
+    return hash.digest('hex');
+  },
+});
+
+export const FILE_METADATA = new TelemetryDataProvider({
+  id: 'appmap.file.metadata',
+  async value({ metadata }: { metadata?: Record<string, unknown> }) {
+    return metadata;
+  },
+});
```
dustinbyrne, line 43: Returning `Record` types is now supported. All the keys of the record will be appended to the telemetry event as `${TelemetryDataProvider.id}.${recordKey}`.

PR #319, opened 2021-08-04
feat: Add `config:write` event

torchello:


src/telemetry/definitions/properties.ts
```diff
@@ -128,3 +138,39 @@ export const PROJECT_LANGUAGE_DISTRIBUTION = new TelemetryDataProvider({
...
+    } catch {
+      return 'false';
+    }
+  },
+});
+
+export const CONFIG_APP = new TelemetryDataProvider({
+  id: 'appmap.config.app',
+  proxyCache: ['agent'],
+  async value({ agent, rootDirectory }: { agent?: AppMapAgent; rootDirectory: PathLike }) {
```
dustinbyrne, line 158: Could this optionally take either:
- a `Uri` of the config so that it may parse the `app` name if `agent` is unavailable
- a JSON representation of the config to do the same
- a `name: string` override, where it'll just return the `name` argument if it's supplied

My concern is that the agent won't be available to provide this information.

src/telemetry/definitions/properties.ts
```diff
@@ -128,3 +138,39 @@ export const PROJECT_LANGUAGE_DISTRIBUTION = new TelemetryDataProvider({
     return JSON.stringify(languageDistribution);
   },
 });
+
+export const CONFIG_IS_VALID = new TelemetryDataProvider({
+  id: 'appmap.config.is_valid',
+  proxyCache: ['agent'],
+  async value({ agent, rootDirectory }: { agent?: AppMapAgent; rootDirectory: PathLike }) {
+    try {
```
dustinbyrne, line 146: If `agent` is `undefined`, return `undefined`.  If the agent is unavailable, we can't reason about whether the file is valid without performing the validation logic ourselves here. Returning `undefined` will omit this property from the telemetry event, instead of reporting an invalid config.

src/telemetry/definitions/properties.ts
```diff
@@ -22,16 +24,24 @@ export const FILE_PATH = new TelemetryDataProvider({
       return undefined;
     }
 
-    return path.relative(rootDirectory as string, uri.fsPath);
+    return fs
```
dustinbyrne, line 27: Prefer `await` with a try/catch. I believe this will return a `Promise<Promise<string>>`, and I'm not certain that'll be properly handled. I wouldn't worry about whether an entity at the given URI exists  or not, report the path nonetheless. For example, if we're describing a file that's deleted, knowing the path of the file is more useful than reporting the path as `deleted`.

src/telemetry/definitions/properties.ts
```diff
@@ -128,3 +138,39 @@ export const PROJECT_LANGUAGE_DISTRIBUTION = new TelemetryDataProvider({
...
+    } catch {
+      return UNKNOWN_APP;
+    }
+  },
+});
+
+export const CONFIG_CONTENT = new TelemetryDataProvider({
+  id: 'appmap.config.content',
+  async value({ uri }: { uri: Uri }) {
+    return fs
```
dustinbyrne, line 171: Prefer `await`. Similar to above, this will return a `Promise<Promise<string>>`, which may not perform as intended. If an error occurs, return `undefined` and omit this property from the telemetry event altogether.

src/telemetry/definitions/properties.ts
```diff
@@ -22,16 +24,24 @@ export const FILE_PATH = new TelemetryDataProvider({
...
   },
 });
 
 export const FILE_SHA_256 = new TelemetryDataProvider({
   id: 'appmap.file.sha_256',
   async value({ uri }: { uri: Uri }) {
-    const hash = createHash('sha256');
-    hash.update(await fs.readFile(uri.fsPath));
-    return hash.digest('hex');
+    return fs
```
dustinbyrne, line 37: Prefer `await`. The `access` call seems redundant, because we're calling `readFile` next, which will raise an exception if the file doesn't exist.

Please return `undefined` on error to omit this property from the telemetry event data instead of reporting an empty string.

src/projectWatcher.ts
```diff
@@ -256,6 +252,20 @@ export default class ProjectWatcher {
...
+          rootDirectory: this.rootDirectory,
+          agent: this.agent,
+          uri: uri,
+        });
+      }
+    };
+
+    configWatcher.onDidCreate(configWatcherListener);
+    configWatcher.onDidChange(configWatcherListener);
+    configWatcher.onDidDelete(configWatcherListener);
```
dustinbyrne, line 268: Delete should send a `project/config:delete` event

dustinbyrne (changes requested):
Looking good! I've added a few minor comments below. 👍

src/projectWatcher.ts
torchello: We'll handle `onDidDelete` separately then.

src/telemetry/definitions/properties.ts
torchello: Updated!

dustinbyrne (approved):

PR #327, opened 2021-08-16
fix: Handle invalid git ignore globs gracefully

dustinbyrne:
If we fail to build a regex from an ignore glob, we now:
- log the glob
- filter the glob out and continue

### DIFF
--- a/src/telemetry/properties/versionControlGit.ts
+++ b/src/telemetry/properties/versionControlGit.ts
@@ -3,17 +3,28 @@ import { PathLike, promises as fs } from 'fs';
 import { join, dirname } from 'path';
 import globToRegExp from 'glob-to-regexp';
 import VersionControlProperties from './versionControl';
+import { Telemetry, DEBUG_EXCEPTION } from '../index';
 
 async function regexFromGitIgnore(filePath: PathLike): Promise<readonly RegExp[]> {
   const directory = dirname(filePath as string);
   const content = await fs.readFile(filePath);
+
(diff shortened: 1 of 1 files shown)


torchello:
Agree with @kgilpin. If don't need this exceptions for analytics (I guess we don't) we can just silently skip.

src/telemetry/properties/versionControlGit.ts
```diff
@@ -3,17 +3,28 @@ import { PathLike, promises as fs } from 'fs';
...
     .split(/^/gm)
     .map((line) => line.trim().replace(/^\//, `${directory}/`))
     .filter((line) => line !== '')
-    .map((glob) => globToRegExp(glob, { flags: 'g' }));
+    .map((glob) => {
+      try {
+        return globToRegExp(glob, { flags: 'g' });
+      } catch {
+        const exception = new Error(`Invalid git ignore pattern: ${glob}`);
+        Telemetry.sendEvent(DEBUG_EXCEPTION, { exception });
```
kgilpin, line 23: Exception should be verified as being the expected type, and not sent to telemetry.

kgilpin (changes requested):

dustinbyrne:
I'd like to know what pattern is raising an exception, because I see no reason we should be failing here. With that data, we can issue a proper fix.

dustinbyrne:
I'll address this in a different manner that replaces `globToRegExp` outright.

Issue #333, opened 2021-08-18
Remove `glob-to-regexp`


dustinbyrne:
Remove the `glob-to-regexp` dependency, custom gitignore logic and replace with [`ignore`](https://www.npmjs.com/package/ignore).

See https://github.com/applandinc/vscode-appland/blob/master/src/telemetry/properties/versionControlGit.ts

PR #334, opened 2021-08-19
fix: remove 'glob-to-regexp' dependency and use 'ignore'

sashahonchar:
Fixes: #333


src/telemetry/properties/versionControlGit.ts
```diff
@@ -41,41 +26,30 @@ async function ignoresForWorkspaceFolder(
...
+    }
   }
 
   public isIgnored(path: PathLike): boolean {
-    if (!this.ignoreRegex) {
-      return false;
-    }
-    const result = this.ignoreRegex.some((re) => re.test(path as string));
-    return result;
+    return this.ignore.ignores(path as string);
```
dustinbyrne, line 53: `path` is guaranteed (?) to be absolute, so this check will always fail. I think we'll need to keep track of the absolute base path of each `.gitignore`, then pick the right `ignore` object to check against.

Example:
Given the following lookup table of git ignores,
``js
{
  '/home/db/dev/applandinc/appmap': ignore(),
  '/home/db/dev/applandinc/appmap/some/directory': ignore(),
}
```

and a call to `isIgnored` with the `path` defined as `/home/db/dev/applandinc/appmap/node_modules`, the first value in the lookup table is used. The basename is subtracted from the path, resulting in `node_modules`, and that value is checked for ignore.

If `path` is `/home/db/dev/applandinc/appmap/some/directory/file.txt`, the second value in the lookup table is chosen because it matches more of the path than the first.

dustinbyrne (changes requested):

src/telemetry/properties/versionControlGit.ts
sashahonchar: Good point, thanks @dustinbyrne 
I made changes and now we'll find all known `.gitignore` files recursively to the root folder and check is needed file ignored in any of those gitignores. https://github.com/applandinc/vscode-appland/pull/334/commits/197531085ba7c3df911213a8ba15d06367741876

dustinbyrne (approved):

PR #353, opened 2021-09-24
fix: Capture debug information on number of workspace folders

dustinbyrne:


src/telemetry/definitions/metrics.ts
```diff
@@ -6,3 +7,10 @@ export const APPMAP_JSON = new TelemetryDataProvider({
     return metrics;
   },
 });
+
+export const NUM_WORKSPACE_FOLDERS = new TelemetryDataProvider({
+  id: 'num_workspace_folders',
+  async value() {
+    return workspace.workspaceFolders?.length || 0;
```
kgilpin, line 14: It seems weird that this could be null.

kgilpin (approved):

src/telemetry/definitions/metrics.ts
dustinbyrne: It's a little weird, but the type requires the null check.
```ts
const workspaceFolders: readonly WorkspaceFolder[] | undefined;
```

PR #367, opened 2021-12-01
feat: JavaScript support in the project picker

### LINKED ISSUE #366: Project Picker recognizes and supports supported Node.js projects
Node.js / mocha / Express.js projects are a perfect match on Unix-like OSes (Node.js version 14+, mocha version 8.0.0+)
Node.js / Express.js projects are a good match on Unix-like OSes

Windows - support coming soon

Supported platforms: https://github.com/applandinc/appmap-agent-js#requirements

dividedmind:
Fixes #366

### DIFF
--- a/package.json
+++ b/package.json
@@ -242,6 +242,7 @@
     "@types/glob": "^7.1.3",
     "@types/mocha": "^8.2.0",
     "@types/node": "^12.19.12",
+    "@types/semver": "^7.3.9",
     "@types/sinon": "^10.0.2",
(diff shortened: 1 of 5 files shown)


dustinbyrne (approved):
Thanks for the quick turn-around on this, it works great! 👍

I think there's a potential enhancement here to look at the `package_lock.json`/`yarn.lock` for express. It could be that they're using a framework like next/nuxt on top of express and this will give a false negative. In any case, this is working as intended and we can leave this as an issue for later.

PR #369, opened 2021-12-08
chore: js+express supported

ptrdvrk:


package.json
```diff
@@ -38,7 +38,9 @@
     "diff",
     "ruby",
     "python",
-    "java"
+    "java",
+    "javascript",
+    "js"
```
apotterri, line 43: If understand this correctly, these are the tags that get displayed here, right?
image

I tried searching for `tag:js`, and got basically useless results. `tag:javascript` was much better, so I wonder if there's value in including `js` here? OTOH, `tag: typescript` *was* useful, so maybe `typescript` should be included in this list?

apotterri (changes requested):

package.json
ptrdvrk: OK, let's change `js` to `typescript`, good call. According to the docs, only 5 tags are permitted but just like many other things on the marketplace, the manifest works with more (mysteriously).

https://code.visualstudio.com/api/references/extension-manifest

apotterri (approved):

PR #384, opened 2022-05-16
feat: Add a Code Objects tree view

kgilpin:
https://us06web.zoom.us/rec/share/BJuAEyykjDyVmInB3DlDDBjIyabRKnOwq_yihUFhKRUIuJbr-hasfi20od-i0rQz.6E7qsESL940QdPo7

See Slack for Passcode. Search: `from:@Brian has:link passcode`


src/services/changeEventDebouncer.ts
```diff
@@ -0,0 +1,25 @@
+import * as vscode from 'vscode';
+
+export default class ChangeEventDebouncer<T> extends vscode.EventEmitter<T> {
```
kgilpin, line 3: @dustinbyrne you may find occasion to use this to prevent a ton of update messages being sent, when one will do.

PR #385, opened 2022-05-19
feat: AppMap usage guide

dustinbyrne:
Running smoke tests:
`yarn tsc && node out/test/smokeTest.js`

There's something odd going on with the clipboard assertions after the most recent nightly release. Clicking the 'copy to clipboard' buttons manually works fine.


kgilpin:
This upgraded installer is very much a step forward, with the indicators and check marks etc

I do want to make sure that we aren’t discarding anything useful in the new version, and I feel like the AppMaps list in the existing install guide is useful - we are skipping over that in the new install guide, but after using it, and having various sessions with users in which they really just want to use AppMap to grok their code, I think it’s a mistake to skip over the AppMaps.

Especially since there are new features in the extension for making AppMap integration with the code much better.

kgilpin (approved):

PR #388, opened 2022-05-27
feat: Indicate whether AppMaps are up-to-date

kgilpin:


src/commands/outOfDateTests.ts
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    vscode.window
+      .showInformationMessage(
+        `${fileLocationsToFilePaths(outOfDate.testLocations).length} test files are out of date`,
+        {
+          modal: false,
+        },
+        { title: TEST_NAMES_LINES, action: copyToClipboard() },
+        { title: TEST_NAMES, action: copyToClipboard(fileLocationsToFilePaths) }
```
dustinbyrne, line 120: What do you think about splitting these two items into separate commands? The first time I'd run this command I hadn't noticed the secondary prompt at the bottom to select an option, and I was left wondering why my clipboard was empty.

src/services/appmapUptodateService.ts
```diff
@@ -0,0 +1,180 @@
...
+  commandArgs: string[];
+  interval?: NodeJS.Timeout;
+  process?: ChildProcess;
+  outofDateAppMapLocations: Set<string> = new Set();
+  outofDateAppMapFiles: Set<string> = new Set();
+
+  constructor(public folder: vscode.WorkspaceFolder) {
+    super();
+
+    this.commandArgs = ['appmap', 'depends'];
```
dustinbyrne, line 28: This is somewhat unrelated, but I noticed this command was outputting `Done in Xs.` due to Yarn v1, causing the service to later log an error:
```
[appmap-uptodate-service] Done in 1.89s. has no indexed metadata (Error: ENOENT: no such file or directory, open '/home/db/dev/applandinc/appland/Done in 1.89s./metadata.json')
```
It's a bit of a non-trivial fix because the extra logging (and the means of disabling it) is dependent on the version of `yarn`. It's not really an issue at the moment but worth making known.

dustinbyrne (approved):
Looks good!

PR #391, opened 2022-06-08
ci: Run system tests in Travis

dustinbyrne:
This is complete, but I'm still observing some flakiness in the tests when running in Travis. It'd be nice to address it, so I'll likely timebox it for a couple of hours.


package.json
```diff
@@ -321,7 +321,8 @@
     "test:unit": "ts-node ./test/unit/*.test.ts",
     "test:extension": "ts-node ./test/integrationTest.ts",
     "test:web-client": "mocha web/test/*.test.mjs",
-    "test": "yarn run test:web-client",
+    "test:system": "node out/test/systemTest.js",
+    "test": "yarn test:web-client && yarn test:system",
```
kgilpin, line 325: Shouldn’t we also run test:extension or are those not working?

dustinbyrne: They're not working, but I could also just skip the tests that aren't passing. Something would be better than nothing. I'll make the change.

PR #392, opened 2022-06-08
fix: Misc service errors

kgilpin:
A couple of issues that I hit doing testing.


kgilpin:
@dividedmind tagging you as well because some of this is due to the little droppings files that the indexer puts in the index directory :-)

PR #397, opened 2022-06-20
Put Python support behind a feature flag

apotterri:
Also, add a fix that seems to have improved the test hit rate (maybe in combination with e4bcbfe). It's still not 100%, though. https://app.travis-ci.com/github/applandinc/vscode-appland/builds/252158020 failed, even though several builds before it were green. When rerun, it passed: https://app.travis-ci.com/github/applandinc/vscode-appland/builds/252158427 .

Fixes https://github.com/applandinc/board/issues/30 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ applications and frameworks!
 
 ## Requirements
 
-Currently, AppMap supports Spring, Django, Flask, Rails and Express projects
+Currently, AppMap supports Spring, Rails, and Express projects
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ applications and frameworks!
 
 ## Requirements
 
-Currently, AppMap supports Spring, Django, Flask, Rails and Express projects
+Currently, AppMap supports Spring, Rails, and Express projects
```
apotterri, line 24: Reviewer(s): do we want this change in the README (that gets pushed to the Marketplace)?

PR #402, opened 2022-06-20
Feature flag python support

apotterri:
Also, add a fix that seems to have improved the test hit rate (maybe in combination with e4bcbfe). It's still not 100%, though. https://app.travis-ci.com/github/applandinc/vscode-appland/builds/252158020 failed, even though several builds before it were green. When rerun, it passed: https://app.travis-ci.com/github/applandinc/vscode-appland/builds/252158427 .

Fixes https://github.com/applandinc/board/issues/30 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ applications and frameworks!
 
 ## Requirements
 
-Currently, AppMap supports Spring, Django, Flask, Rails and Express projects
+Currently, AppMap supports Spring, Rails, and Express projects
 ([AppMap for JavaScript](https://appland.com/docs/reference/appmap-javascript.html) is now available
 in beta!). The most up-to-date list of frameworks and languages we support is
 [here](https://appland.com/docs/integrations.html).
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ applications and frameworks!
 
 ## Requirements
 
-Currently, AppMap supports Spring, Django, Flask, Rails and Express projects
+Currently, AppMap supports Spring, Rails, and Express projects
```
apotterri, line 24: Reviewer(s): do we want this change in the README (that gets pushed to the Marketplace)?

kgilpin: Sure

src/services/languageResolver.ts
```diff
@@ -263,7 +264,8 @@ export default class LanguageResolver {
   private static async identifyLanguage(rootDirectory: PathLike): Promise<string> {
     const languages = await this.getLanguageDistribution(rootDirectory);
     const best = Object.entries(languages).sort((a, b) => b[1] - a[1])?.[0]?.[0];
-    if (LANGUAGE_AGENTS[best]) return best;
+    const agent = LANGUAGE_AGENTS[best];
+    if (agent && agent.enabled) return best;
     return UNKNOWN_LANGUAGE;
```
kgilpin, line 269: I guess it would be a little better if we indicated that the language is Known but Unsupported.

PR #406, opened 2022-06-21
feat: Run background node processes with Electron node

dustinbyrne:
This modifies the background services to run as follows:
- On extension startup, pull `yarn` dependencies from npm if need be (i.e., install `@appland/appmap`, `@appland/scanner`)
- Run yarn using Node shipped with Electron (via the `code` binary) to launch node services

More information is documented [here](https://gist.github.com/dustinbyrne/b4ddea558b6efa3f031ca343aa9251e7).

Closes applandinc/board#34


src/extension.ts
```diff
@@ -17,8 +17,6 @@ import extensionSettings from './configuration/extensionSettings';
 import { FindingsTreeDataProvider } from './tree/findingsTreeDataProvider';
 import WorkspaceServices from './services/workspaceServices';
 import { AppMapWatcher } from './services/appmapWatcher';
-import AutoIndexerService from './services/autoIndexer';
-import AutoScannerService from './services/autoScanner';
```
kgilpin, line 21: So should these files be removed from Git then? It looks like they aren't.

src/configuration/extensionSettings.ts
```diff
@@ -29,22 +29,13 @@ export default {
     ),
 
   indexCommand: (): string | string[] =>
-    vscode.workspace.getConfiguration('appMap').get('indexCommand') || [
-      'appmap',
-      'index',
-      '--watch',
-    ],
+    vscode.workspace.getConfiguration('appMap').get('indexCommand') || ['index', '--watch'],
```
kgilpin, line 32: I tested changing this command setting to `foobar`. What I observed is that I didn't get an error about the command failing. The code as changed in this version doesn't split the command string, so that may be another problem. Maybe we should get rid of these custom settings now? The reason they exist was to give us a way to troubleshoot the commands if they weren't working through `yarn` or `npm`. I guess we don't need that any more.

dustinbyrne: I've dropped the configurable index/scan/depends commands

src/extension.ts
```diff
@@ -268,6 +250,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
 
     registerUtilityCommands(context, extensionState);
 
+    // The node dependencies may take some time to retrieve. As a result, the initialization sequence is
+    // wrapped in an async function but we won't wait for it to resolve.
+    (async function() {
```
kgilpin, line 255: I think I'm observing a problem with `test:extension`. The appmap process services wants to resolve the binary e.g. `appmap`, but that binary isn't installed until this installer is finished. So the process service raises an exception from 

```
    const binPath = await getBinPath(options.bin);
```

kgilpin (approved):

PR #409, opened 2022-06-22
fix: Create getTreeName function to format tree name

MaxDillon:
[Issue 64](https://github.com/applandinc/board/issues/64)

This PR changes the way items are listed in the Findings menu in VSCode.

New Pattern: `[Rule] in [context], [file:line]`

### DIFF
--- a/src/lib/generateDisplayTitle.ts
+++ b/src/lib/generateDisplayTitle.ts
@@ -1,0 +1,20 @@
+import { ResolvedFinding } from '../services/resolvedFinding';
+import * as vscode from 'vscode';
+
+// Gets's name displayed in Findings bar
+export default (finding: ResolvedFinding): string => {
+  const absPath = finding.problemLocation?.uri.path;
+  const relPath = absPath ? vscode.workspace.asRelativePath(absPath) : undefined;
+
+  const rule = finding.finding.ruleTitle;
+  const context = finding.finding.groupMessage || finding.finding.message;
+  const lineno = finding.problemLocation?.range.start.line;
+
+  // If rule and context are the same, only display one
+  const ruleAndContext = rule !== context ? `${rule}: ${context}` : rule;
+
+  // Only display problemLocation if it exists
+  const fullPathString = finding.problemLocation ? `, ${relPath}:${lineno}` : '';
+
+  return `${ruleAndContext}${fullPathString}`;
+};
--- a/src/tree/findingsTreeDataProvider.ts
+++ b/src/tree/findingsTreeDataProvider.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 import FindingsIndex from '../services/findingsIndex';
 import { ResolvedFinding } from '../services/resolvedFinding';
+import generateTitle from '../lib/generateDisplayTitle';
 
 export class FindingsTreeDataProvider implements vscode.TreeDataProvider<vscode.TreeItem> {
   private _onDidChangeTreeData = new vscode.EventEmitter<undefined>();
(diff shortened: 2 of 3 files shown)


src/tree/findingsTreeDataProvider.ts
```diff
@@ -2,6 +2,11 @@ import { Event } from '@appland/models';
 import * as vscode from 'vscode';
 import FindingsIndex from '../services/findingsIndex';
 import { ResolvedFinding } from '../services/resolvedFinding';
+import { Finding } from '@appland/scanner/built/cli';
+
+function getTreeName(finding: Finding) {
+  return `${finding.ruleTitle}: ${finding.groupMessage || finding.message}, ${finding.stack[0]}`;
```
kgilpin, line 8: If you dig into the code that opens the source code link, we don’t always link to the first stack entry. We will do that unless we can find a relative path to a source file, which we assume is part of the project source and therefore something the developer has more control over.

MaxDillon: There was another property of the findings object called scope:
```
  "scope": {
    "static": true,
    "id": 1,
    "event": "call",
    "thread_id": 1,
    "defined_class": "misago.faker.tests.test_fake_categories",
    "method_id": "test_fake_category_can_be_created_with_copy_of_other_category_acls",
    "path": "misago/faker/tests/test_fake_categories.py",
    "lineno": 18
  },
```
I initially didn't use it because it wassn't in all my test findings. Would this be the best path to use? Then I could fallback on the stack?

kgilpin:
Scope is more like the most top level event within which the problem
occurs.

I’m not sure which is the best. We would probably need to look at some
examples.

kgilpin:
ResolvedFinding has the location you need.

src/tree/findingsTreeDataProvider.ts
MaxDillon: There are some examples like this
```
  "scope": {
    "http_server_request": {
      "request_method": "POST",
      "protocol": "HTTP/1.1",
      "path_info": "/admincp/graphql/",
      "normalized_path_info": "/admincp/graphql/",
      "mime_type": "application/json",
      "headers": {
        "Cookie": "sessionid=bkydqv3eodpvl0zp98t64cuguyx357vh",
        "Content-Length": "776",
        "Content-Type": "application/json"
      }
    },
```
Here there is an HTTP request instead of a path. 
- I could use another method to get the path in this case
- I could use the HTTP Request path instead: e.g. "/admincp/graphql/"

kgilpin:
Use ResolvedFinding.problemLocation ; it's there for this

https://github.com/applandinc/vscode-appland/blob/dd4004269f5c452a58064debf11639eb0c149ec3/src/services/resolvedFinding.ts#L39

src/tree/findingsTreeDataProvider.ts
dustinbyrne: The scope is essentially the root location of the finding. For example when searching for N+1 queries we only consider queries which originate from a request handler (i.e. an event containing `http_server_request`). So while the scope will reflect the origin request, the actual call site of the N+1 query will have occurred somewhere deeper in the call stack. If we're looking to identify the **file** containing the problem, we'd likely want to use a path from the `event` or `relatedEvents`.

src/lib/generateDisplayTitle.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// Gets's name displayed in Findings bar
+export default (finding: ResolvedFinding): string => {
+  const absPath = finding.problemLocation?.uri.path;
+  const relPath = absPath ? vscode.workspace.asRelativePath(absPath) : undefined;
+
+  const rule = finding.finding.ruleTitle;
+  const context = finding.finding.groupMessage || finding.finding.message;
+  const path = finding.finding.scope.path || relPath;
+  const lineno = finding.finding.scope.lineno || finding.problemLocation?.range.start.line;
```
dustinbyrne, line 12: I think our best bet is to just drop the scope entirely. The same finding (hash) may appear in many different scopes.

dustinbyrne: If there's no `problemLocation` we can just omit the path for now

MaxDillon: Ok, done

MaxDillon:
### Screenshots
Screen Shot 2022-06-24 at 10 16 19 AM
Screen Shot 2022-06-24 at 10 16 37 AM

dustinbyrne (approved):

PR #410, opened 2022-06-23
fix: Pass disabledLanguages to InstallGuide

apotterri:
When opening the InstallGuide, pass disabledLanguages so it can be made
available to nested components.

### DIFF
--- a/web/src/installGuideView.js
+++ b/web/src/installGuideView.js
@@ -7,55 +7,59 @@ export default function mountInstallGuide() {
   const vscode = window.acquireVsCodeApi();
   const messages = new MessagePublisher(vscode);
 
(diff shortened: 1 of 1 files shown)


apotterri:
How are the Vue components getting access to the information from the analyzer in `python.disabled.ts`? The analyzer is in the extension, but the Vue components are in a dependency (`@appland/components`).

ahtrotta:
Sorry for being unclear. I haven't followed the entire execution chain, but the data from the analyzers eventually ends up in the `projects` prop (line 43 of `InstallGuide.vue`). I found that when I was rendering the project picker, the data in `projects` was sufficient—I didn't end up using the `disabledLanguages` prop.

ahtrotta:
Yup! I'll remove `disabledLanguages` 👍

PR #417, opened 2022-06-27
ci: Run extension tests

kgilpin:


package.json
```diff
@@ -322,7 +322,7 @@
     "test:extension": "ts-node ./test/integrationTest.ts",
     "test:web-client": "mocha web/test/*.test.mjs",
     "test:system": "node out/test/systemTest.js",
-    "test": "yarn test:web-client && yarn test:system",
+    "test": "yarn test:web-client ; yarn test:extension ; yarn test:system",
```
apotterri, line 325: Switching from `&&` to `;` here will run all three kinds of tests, but only failures in `test:system` will cause the build to fail. It seems like it would be better to use `&&`.

apotterri (changes requested):

kgilpin:
Yeah I just wanted them all to run for now while we are trouble shooting

PR #421, opened 2022-06-28
fix: Fix fqid conflicts in the Code Objects view

kgilpin:
Fix a class map merging problem that caused duplicate ids in the Code Objects view.

This is caused by the behavior of appmap-ruby to emit package names like `some/package`, rather than splitting these into a parent and child.


test/integration/buildClassMap.test.ts
```diff
@@ -0,0 +1,242 @@
...
+  package:app/services
+    class:app/services/ApplicationService
+      function:app/services/ApplicationService#client=
+    class:app/services/ApplicationServiceClient
+      function:app/services/ApplicationServiceClient#client
+    class:app/services/GitHub
+      class:app/services/GitHub::Clients
+        class:app/services/GitHub::Clients::App
+          function:app/services/GitHub::Clients::App#authenticate_installation
+    package:app/services/git_hub
```
kgilpin, line 200: This is the key test, ensuring that `git_hub/clients` only occurs once uniquely in the code objects tree.

src/services/classMapIndex.ts
```diff
@@ -276,126 +61,3 @@ export default class ClassMapIndex {
     }
   }
 }
-
-async function buildClassMap(
```
kgilpin, line 280: It made more sense to test this as a separate function, decoupled from this service.

src/lib/buildClassMap.ts
```diff
@@ -0,0 +1,142 @@
...
+    cme.children = cme.children || [];
+    const betterName = betterNames[[cme.type, cme.name].join(' ')];
+    if (betterName) cme.name = betterName;
+
+    let result = cme;
+    if (cme.type === CodeObjectEntryRootType.PACKAGE) {
+      {
+        // Some code object entries have a path-delimited package name, but we want
+        // each package name token to be its own object.
+        const chain = cme.name.split('/').map(
```
kgilpin, line 42: It's important that this logic be applied to all PACKAGEs as they are merged into the tree.

src/lib/CodeObjectEntry.ts
```diff
@@ -0,0 +1,208 @@
...
+    );
+  }
+
+  merge(appMapFilePath: string, other: ClassMapEntry): void {
+    if (other.labels) {
+      for (const label of other.labels) {
+        this.labels.add(label);
+      }
+    }
+    this._appMapFiles.add(appMapFilePath);
```
kgilpin, line 168: No need to add in the children here, that will be done later.

test/integration/buildClassMap.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        !(coe.type === CodeObjectEntryRootType.PACKAGE && coe.name.split('/').length > 1),
+        `Expected ${coe.fqid} ${coe.name} to have exactly one token`
+      );
+      coe.children.forEach((child) => hasOneToken(child));
+    };
+    classMap.forEach((coe) => hasOneToken(coe));
+
+    assert.deepStrictEqual(
+      classMapDescription,
+      `package:actionpack
```
apotterri, line 38: At some point, maybe this (and other enormous strings) should get pushed out into separate files. It would make the test easier to read.

test/integrationTest.ts
```diff
@@ -154,4 +156,9 @@ const testWorkspaces = [PROJECT_A, PROJECT_UPTODATE];
     }
   }
   process.exit(succeeded ? 0 : 1);
-})();
+}
+
+integrationTest().catch((e) => {
+  console.warn(e);
+  process.exit(1);
```
apotterri, line 163: As long as you're here, we should consider changing this to set `process.exitCode`, rather than calling `process.exit`.

While I was investigating some of the build failures, I noticed that some child processes weren't getting cleaned up correctly. This problem was masked by calling `process.exit`: the process would exit, but the children would linger on, now owned by PID 1. If we just set `process.exitCode`, the process will hang at exit, and we'll know we have an issue.

(The changes you and DB made in the Electron node branch fixed the problem I was seeing, and as far as I can tell, there are no longer issues cleaning up any children that get created.)

src/lib/CodeObjectEntry.ts
```diff
@@ -0,0 +1,208 @@
+import * as vscode from 'vscode';
```
apotterri, line 1: This code just got refactored into its own file, yes? Just wanted to confirm that there are no changes to functionality that need to be reviewed.

apotterri (changes requested):

src/lib/CodeObjectEntry.ts
kgilpin: There are two changes, really. One is not requiring the WorkspaceFolder any more. The other is not merging the children here, because it's actually done later, and properly, by `mergeCodeObject`:

```typescript
  const mergeCodeObject = (
    folder: vscode.WorkspaceFolder,
    appMapFilePath: string,
    parent: CodeObjectEntry,
    cme: ClassMapEntry
  ) => {
    cme = normalizeClassMapEntry(cme);
    const child = parent.findOrMergeChild(folder, appMapFilePath, cme);
    codeObjectByFqid[child.fqid] = child;
    (cme.children || []).forEach(mergeCodeObject.bind(null, folder, appMapFilePath, child));
  };
```

Screen Shot 2022-06-30 at 9 50 01 AM

apotterri: Ok, thanks. FWIW, I've been trying to use `refactor` commits to move code around, then separate `fix` or `feat` commits when I make changes.

kgilpin:
Yeah good idea.

kgilpin:
@apotterri I've made some changes and all the extension tests are passing locally.

apotterri (approved):
Looks good, thanks.

PR #422, opened 2022-06-28
Fix: show python and js as open beta

ahtrotta:
Fixes applandinc/board#86

Change messages that get displayed in project picker. Show that Python and JavaScript are in Open Beta. Remove Python feature flag.


brikelly (commented):
I haven't vetted the code or behavior, but the message strings all look like good edits.

apotterri (changes requested):
Code changes look good here. Behavior should get tested by the new project picker tests, right @ahtrotta?

This needs a rebase before it can be merged. If you do it as an interactive rebase, you can reword the commit subject lines to include a commit type: 
image

apotterri (approved):

PR #423, opened 2022-06-29
Fix: update scoring in the language analyzers

ahtrotta:
Fixes: applandinc/board#91

When language, web framework, and test framework are all yellow ('ok') then the project picker says `This project meets all requirements necessary to create AppMaps.` in green.

Before:
Image

After:
image

Also, when there is an unsupported web framework, the project picker says that you can continue:

Before:
Image

After:
image

### DIFF
--- a/src/analyzers/index.ts
+++ b/src/analyzers/index.ts
@@ -5,9 +5,25 @@ import LanguageResolver from '../services/languageResolver';
 
 export type Score = 'bad' | 'ok' | 'good';
 const SCORE_VALUES = { bad: 0, ok: 1, good: 2 };
+const OVERALL_SCORE_VALUES = { bad: 1, ok: 2, good: 3 };
 
 export function scoreValue(...scores: Score[]): number {
   return scores.reduce((s, x) => s + SCORE_VALUES[x], 0);
+}
+
+export function overallScore(result: Features): number {
+  const languageScore: string = result.lang.score;
+  const webFrameworkScore: string = result?.web?.score || 'bad';
+  const testFrameworkScore: string = result?.test?.score || 'bad';
+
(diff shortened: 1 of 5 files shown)


src/analyzers/index.ts
```diff
@@ -5,11 +5,27 @@ import LanguageResolver from '../services/languageResolver';
...
+const OVERALL_SCORE_VALUES = { bad: 1, ok: 2, good: 3 };
 
 export function scoreValue(...scores: Score[]): number {
   return scores.reduce((s, x) => s + SCORE_VALUES[x], 0);
 }
 
+export function overallScore(result: Features): number {
+  const languageScore: string = result.lang.score;
+  const webFrameworkScore: string = result?.web?.score || 'bad';
+  const testFrameworkScore: string = result?.test?.score || 'bad';
```
dividedmind, line 17: These aren't `string`s but `Score`s. In fact, I'm not sure you need to explicitly type them at all; the `Score` type should be inferred from the initializer.

dividedmind (approved):
Thanks for fixing this, it definitely better matches our intentions now!

I left a minor tip for you re typing; I leave this up to you if you want to look into it before merging. Another thing I'd suggest is checking if a unit test can be added. However, none of these is blocking the merge IMO.

One thing that you need to do before merging is squashing the two commits together and using a commit title that semantic-release can recognize; you can add a longer sorta-user-visible description in the commit text body. See https://github.com/applandinc/vscode-appland/commit/2aad9cae9ec7dbd2f55e255489c115b546fdbbad for an example of how I write these, but the important thing is the `fix: ` (note the lower case; I know, it bugs me too ;) ) prefix in the commit title.

Thanks!

PR #424, opened 2022-06-29
Fix: include whether user has a supported node version in project metadata

ahtrotta:
Fixes applandinc/board#88

This adds a boolean to the project metadata that signifies if a supported node version was detected. This gets propagated to the frontend so that we can conditionally render a message helping users install the correct version of node.


ahtrotta:
Created a new branch to clean up commit history.

PR #425, opened 2022-06-30
Fix: send node status to frontend

ahtrotta:
Fixes https://github.com/applandinc/board/issues/88

This adds a boolean to the project metadata that signifies if a supported node version was detected. This gets propagated to the frontend so that we can conditionally render a message helping users install the correct version of node.


apotterri (approved):
Let's not forget to add some tests for this.

PR #429, opened 2022-06-30
fix: Tolerate partial node version in .nvmrc

kgilpin:


src/analyzers/index.ts
```diff
@@ -103,31 +103,27 @@ export async function analyze(
...
+    .map(Number);
 
-  if (digits.length !== 3) {
-    return {
-      major: 0,
-      minor: 0,
-      patch: 0,
-    };
-  }
+  while (digits.length < 3) digits.push(0);
```
apotterri, line 126: This is definitely better than it was. Having `16` in your `.nvmrc` means "use the latest version", though, doesn't it, rather than "use 16.0.0"? Probably not import for the moment....

apotterri (approved):

kgilpin:
Yeah it does. We only check major version anyway - and probably only ever
will...

PR #433, opened 2022-07-07
Fix/investigate findings green check

ahtrotta:
Fixes applandinc/board#57

The green check mark next to `Investigate findings` now appears when a user presses the 'View problems' button or when a user clicks on a tree item in the `Runtime Analysis` tree view.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,7 +6,7 @@ import registerTrees from './tree';
 import AppMapCollectionFile from './services/appmapCollectionFile';
 import ContextMenu from './tree/contextMenu';
 import RemoteRecording from './actions/remoteRecording';
-import { notEmpty } from './util';
+import { getWorkspaceFolderFromPath, notEmpty } from './util';
 import { registerUtilityCommands } from './registerUtilityCommands';
 import ExtensionState from './configuration/extensionState';
 import FindingsIndex from './services/findingsIndex';
@@ -41,6 +41,7 @@ import assert from 'assert';
 import { initializeWorkspaceServices } from './services/workspaceServices';
 import { NodeProcessService } from './services/nodeProcessService';
 import getEarlyAccess from './commands/getEarlyAccess';
+import openFinding from './commands/openFinding';
 import { RuntimeAnalysisCtaService } from './services/runtimeAnalysisCtaService';
 
 export async function activate(context: vscode.ExtensionContext): Promise<AppMapService> {
@@ -183,7 +184,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
       badge.initialize(projectStates);
       context.subscriptions.push(badge);
 
-      InstallGuideWebView.register(context, projectStates);
+      InstallGuideWebView.register(context, projectStates, extensionState);
       InstallGuideWebView.tryOpen(extensionState);
 
       processService = new NodeProcessService(context, projectStates);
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -134,6 +134,10 @@
         "command": "appmap.copyOutOfDateTestsToClipboard",
         "title": "AppMap: Copy Out-of-Date Tests to Clipboard"
       },
+      {
+        "command": "appmap.openFinding",
+        "title": "AppMap: Open Finding"
+      },
       {
```
dustinbyrne, line 141: This command doesn't look to be intended to be used by an end-user, so we shouldn't include it here. The command definitions here are added to the command palette. It's not necessary to add an entry here when creating a new command via the API (i.e. `vscode.commands.registerCommand`).

src/util.ts
```diff
@@ -244,6 +245,17 @@ export function hasPreviouslyInstalledExtension(extensionPath: string): boolean
...
+  projectStates: ProjectStateServiceInstance[],
+  path: string
+): Promise<vscode.WorkspaceFolder | undefined> {
+  const projectState = projectStates.find(async (projectState) => {
+    return (await projectState.metadata())?.path === path;
+  });
+
+  return projectState?.folder;
+}
+
```
dustinbyrne, line 258: I'm not sure this is working as intended.

`path` will be something like `file:///home/user/my-project/app/controllers/posts_controller.rb`. It will never be equal to the project path. This is only working because upon hitting the first `await`, a promise is returned and `find` will return the first element in the `projectStates` array due to the truthy return value (a promise).

ahtrotta: Interestingly, they are the same in my environment. But if they're not the same on Mac or Windows, then I definitely need to change this logic. Perhaps I should use a RegEx match?

image

ahtrotta: Ah, I discovered that they're the same when I click on the `View problems` button, but not when I click on a findings Tree item. I'll fix this up.

ahtrotta: Okay, so I revamped this function. I modified it so that it uses `Promise.all` to avoid always returning the first element in the `projectStates` array. I also check if the `path` parameter includes the the `path` from the `projectState`, because something like `file:///home/user/my-project/app/controllers/posts_controller.rb` will contain `/home/user/my-project/` as a substring.

src/extension.ts
```diff
@@ -203,6 +203,16 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
...
 
+    if (findingsEnabled) {
+      vscode.commands.registerCommand('appmap.openFinding', async (uri: vscode.Uri) => {
+        const workspaceFolder = await getWorkspaceFolderFromPath(projectStates, String(uri));
+        if (workspaceFolder) {
+          extensionState.setFindingsInvestigated(workspaceFolder, true);
+        }
+        vscode.commands.executeCommand('vscode.open', uri);
+      });
+    }
```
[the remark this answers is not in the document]
ahtrotta: Yup, good idea

apotterri (changes requested):
It doesn't seem like the commits in this PR should be separate -- the same files are touched by many of them, and the comments only have (not very helpful) subject lines.

If they're all work towards the same end-goal, they should get combined into a single commit with a descriptive comment. (i.e. don't just squash them and that the subjects). If they got split up as a result of evolution during the review process, you should consider using [fixup commits](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---fixupamendrewordltcommitgt). These make it easy to clean things up using the [auto-squash](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash) switch when doing a final rebase.

kgilpin:
Seconded

ahtrotta:
Okay, I squashed all of the commits into one commit and tried (hopefully successfully, but let me know if not) to make a better commit message

apotterri (changes requested):
This looks much better, thanks. A couple more minor changes:

* Because the comments on a PR more or less disappear once it gets merged, this should go on the commit, rather than the PR:
> The green check mark next to Investigate findings now appears when a user presses the 'View problems' button or when a user clicks on a tree item in the Runtime Analysis tree view.

* the subject line of the commit should be shorter than 72 characters, ideally [not more than 50](https://cbea.ms/git-commit/#limit-50):
image

ahtrotta:
I believe I addressed your comment, but let me know if it's not what you were looking for

apotterri (approved):

PR #435, opened 2022-07-08
feat: send findings domain counts to frontend webview

ahtrotta:
Fixes applandinc/board#28

Makes the number of each type of finding available to the frontend


src/extension.ts
```diff
@@ -167,8 +167,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
 
     let processService: NodeProcessService | undefined;
     let runtimeAnalysisCta: RuntimeAnalysisCtaService | undefined;
+    let projectState: ProjectStateService;
     {
-      const projectState = new ProjectStateService(
+      projectState = new ProjectStateService(
```
dustinbyrne, line 172: If we need `projectState` available to the outer scope, we can just drop the scope it's defined in.

src/appMapService.ts
```diff
@@ -7,6 +7,7 @@ import { AppMapConfigWatcher } from './services/appMapConfigWatcher';
 import { AppmapUptodateService } from './services/appmapUptodateService';
 import Command from './services/command';
 import { NodeProcessService } from './services/nodeProcessService';
+import ProjectStateService, { ProjectStateServiceInstance } from './services/projectStateService';
```
dustinbyrne, line 10: `ProjectStateServiceInstance` looks to be unused

src/extension.ts
ahtrotta: Okay, I dropped the scope. Why was it there in the first place?

src/appMapService.ts
ahtrotta: I deleted it

test/fixtures/workspaces/project-several-findings/.nvmrc
```diff
@@ -0,0 +1 @@
+v16.15.0
```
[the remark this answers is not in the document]
ahtrotta: Deleted!

dustinbyrne (approved):

PR #437, opened 2022-07-13
Fix the system tests

### LINKED ISSUE #399: VS Code system tests are failing intermittently

apotterri:
See the comment on [8dc414e](https://github.com/applandinc/vscode-appland/pull/437/commits/8dc414e51ee7ee7acf858a203fa1cf13ecb2d2be).

Fixes #399 .


test/system/tests/instructions.test.ts
```diff
@@ -114,20 +126,26 @@ describe('Instructions tree view', function() {
     await assertTabs(1);
   });
 
-  it('hides the pending badge upon completing the installation steps', async () => {
+  xit('hides the pending badge upon completing the installation steps', async function() {
```
apotterri, line 129: This test is still being skipped, not so much because it's failing, but more because I couldn't figure out what it's supposed to be testing.

@ahtrotta Can you help me figure out what this is supposed to be doing, please?

ahtrotta: Yeah, the `pending badge` is the little blue clock icon that is displayed on the AppMap extension icon. See here:
image

This test ensures that it disappears once a user completes all the instruction steps:
image

apotterri: Ah, ok, I understand now. I'll figure out how to make the test pass.

apotterri: So, it looks like there are a couple of problems here: 
* the second call to `appMap.openActionPanel` closes it (suggesting that's it's name should actually be `toggleActionPanel`)
* the pending badge goes away when `appmap-findings.json` gets restored:
image

The first is easily addressed, of course. But, is the second correct, or should the pending badge remain until the user has viewed the problem?

apotterri: Hmmm, ok. It's definitely not working that way. So, it seems like there's really a problem here, but not one related to the reliability of the tests. We should open a separate issue to track it.

test/systemTest.ts
```diff
@@ -51,29 +47,45 @@ async function main(): Promise<void> {
...
+      const driver = new Driver(app, context, page);
+      context.setDefaultTimeout(TIMEOUT);
+
+      await driver.waitForReady();
+      this.app = app;
+      this.driver = driver;
+      this.project = project;
+    };
+    const afterEach: Mocha.Func = async function() {
+      await Promise.all([this.app.waitForEvent('close'), await this.app.close()]);
```
ahtrotta, line 75: You don't need to use `await` inside of `Promise.all` since it's expecting an array of Promises.

test/unit/project.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      it('can filter out AppMaps', async function() {
+        await this.projectDirectory.reset('**/*.appmap.json');
+        expect(this.jsonPath).not.to.be.a.path();
+      });
+
+      it('can filter out the config file', async function() {
+        const dir = await this.projectDirectory.reset('appmap.yml');
+        expect(join(dir.workspacePath, 'appmap.yml')).not.to.be.a.path();
+      });
```
ahtrotta, line 45: Missing an empty line after this line.

test/system/tests/instructions.test.ts
```diff
@@ -1,30 +1,34 @@
...
-  beforeEach(async () => {
     await driver.closePanel();
-    await project.reset();
-    await project.removeAppMapFiles();
     await driver.resetUsage();
     await driver.reload();
   });
 
-  it('displays a badge when AppMap installation is pending for an installable project', async () => {
+  it('displays a badge when AppMap installation is pending for an installable project', async function() {
```
ahtrotta, line 19: Why did you switch from arrow functions to function declarations? (I'm not implying that you have to switch back to arrow functions. I just want to know for future reference.).

ahtrotta (changes requested):
This looks good! There are only a few minor changes that I'd like to see (see comments). Definitely let me know if any of my comments are confusing or reveal that I'm misunderstanding something.

test/system/tests/instructions.test.ts
apotterri: I switched so I could use Mocha's context to store the driver and project. The context isn't available if you use arrow functions: https://mochajs.org/#arrow-functions .

extern/yarn.lock
```diff
@@ -102,17 +102,17 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@appland/components@npm:2.5.0":
-  version: 2.5.0
-  resolution: "@appland/components@npm:2.5.0"
+"@appland/components@npm:2.7.0":
+  version: 2.7.0
+  resolution: "@appland/components@npm:2.7.0"
```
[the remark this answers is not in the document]
apotterri: Yarn is choosing the version of `@appland/components` here, based on the requirements of `@appland/appmap`. When `@appland/appmap` gets updated, we'll want to update the extension, too. (It's never, ever a good idea to update a lock file manually. You always want to let the package manager do it.)

test/system/src/app.ts
```diff
@@ -28,17 +28,26 @@ interface LaunchOptions {
...
   options: LaunchOptions
 ): Promise<CodeHarness> {
-  const codePath = await downloadAndUnzipVSCode('insiders');
   const baseCodePath = path.join(codePath, '..');
   const executablePath = await getElectronPath(baseCodePath);
   const app = await Electron.launch({
     executablePath,
     args: [
+      process.env.INSPECT_EXTENSION ? '--inspect-brk-extensions' : '',
+      process.env.INSPECT_EXTENSION ? '9229' : '',
```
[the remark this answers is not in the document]
apotterri: Passing `--inspect-brk-extensions` when launching vscode is supposed to be a way to debug an extension. Port `9229` is the default port the debugger will connect to.

This appears to be broken in the latest Insider's build, though, so I've taken this setting out.

test/system/tests/instructions.test.ts
```diff
@@ -70,9 +77,16 @@ describe('Instructions tree view', function() {
       InstructionStep.OpenAppMaps,
       InstructionStepStatus.Complete
     );
+
+    await driver.appMap.assertInstructionStepStatus(
```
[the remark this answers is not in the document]
apotterri: Fixed.

test/systemTest.ts
```diff
@@ -20,17 +21,12 @@ async function main(): Promise<void> {
...
 
     let verbose = false;
-    let leaveOpen = false;
     let testPath = path.join(__dirname, 'system', 'tests', '**', '*.test.js');
     const patterns: string[] = [];
     const cliArgs = process.argv.slice(2);
     while (cliArgs.length > 0) {
       const arg = cliArgs.shift() as string;
       switch (arg) {
-        case '--leave-open':
```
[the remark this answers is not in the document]
apotterri: There is now a new instance of vscode being launched for every test case: https://github.com/applandinc/vscode-appland/pull/437/files#diff-40de7231596e00146a30b40f1ba91071ba798e4c4f1dd588e966546690bf4a84R57, so this switch no longer makes sense. To see what's happening in a test, you can run with a debugger attached.

test/unit/project.test.ts
apotterri: Fixed.

test/system/tests/support/project.ts
```diff
@@ -1,78 +1,80 @@
...
+  private projectPath: string;
+  private projectName: string;
+  public readonly workspacePath: string;
 
-const ignoreExceptions = async (fn: (...args: any[]) => any) => {
-  try {
-    return await fn();
-  } catch {
-    // Do nothing
+  // Prepare a new directory to become a copy the given project. It will be empty until reset is called.
```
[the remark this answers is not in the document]
apotterri: Fixed.

test/systemTest.ts
apotterri: Fixed.

test/system/tests/instructions.test.ts
apotterri: @ahtrotta Can you open an issue for this, please?

ahtrotta: Yup, I'll do that.

test/system/tests/instructions.test.ts
```diff
@@ -179,7 +197,7 @@ describe('Instructions tree view', function() {
     //   );
     // }
 
-    await project.restoreFile('*.appmap.json');
+    await project.restoreFiles('**/*.appmap.json');
     await driver.instructionsWebview.clickButton('Next');
     await driver.instructionsWebview.getPageByTitle('Open AppMaps').waitFor();
```
dustinbyrne, line 202: Since the rebase, the `Open AppMaps` title in the webview appears to have changed to `Explore AppMaps`, which will cause this to fail. This is referenced in at least a couple of different areas.

dustinbyrne:
The build is passing but there are four tests failing, presumably due to a rebase? There's an `|| true` in `.travis.yml` allowing the build to pass regardless of the test results. We can probably remove that now.

test/system/tests/instructions.test.ts
```diff
@@ -48,31 +53,35 @@ describe('Instructions tree view', function() {
...
 
     await driver.appMap.assertInstructionStepStatus(
       InstructionStep.RecordAppMaps,
       InstructionStepStatus.Complete
     );
 
-    await project.restoreFile('**/appmap-findings.json');
+    await project.restoreFiles('**/appmap-findings.json');
     await driver.appMap.openInstruction(InstructionStep.InvestigateFindings);
     await driver.instructionsWebview.clickButton('View problems');
```
ahtrotta, line 68: This button has changed and the label is now `'Open the PROBLEMS tab'`.

test/system/tests/instructions.test.ts
```diff
@@ -114,20 +125,25 @@ describe('Instructions tree view', function() {
...
+    await driver.waitForFile(pidfile);
     await driver.appMap.openActionPanel();
-    await project.restoreFile('*.appmap.json');
+    await project.restoreFiles('**/*.appmap.json');
     await driver.appMap.openAppMap();
-    await project.restoreFile('**/appmap-findings.json');
-    await driver.appMap.openActionPanel();
+    await project.restoreFiles('**/appmap-findings.json');
     await driver.appMap.openInstruction(InstructionStep.InvestigateFindings);
     await driver.instructionsWebview.clickButton('View problems');
```
ahtrotta, line 140: This button has changed and the label is now 'Open the PROBLEMS tab'.

test/system/tests/instructions.test.ts
apotterri: Fixed.

test/system/tests/instructions.test.ts
apotterri: Fixed.

test/system/tests/instructions.test.ts
apotterri: Fixed.

apotterri:
> The build is passing but there are four tests failing, presumably due to a rebase? There's an `|| true` in `.travis.yml` allowing the build to pass regardless of the test results. We can probably remove that now.

I've rearranged this a bit, so we're only using `|| true` for `test:extension`. These tests still fail intermittently.

ahtrotta (approved):

PR #451, opened 2022-08-06
feat: Keep viewer state when switching tabs

dividedmind:
One annoying aspect of the editor is that when switching tabs the current context is lost. This fixes this problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -363,6 +363,7 @@
     "openapi-types": "^11.0.1",
     "prettier": "^1.19.1",
     "prettier-eslint": "^12.0.0",
+    "project-root-directory": "^1.0.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -363,6 +363,7 @@
     "openapi-types": "^11.0.1",
     "prettier": "^1.19.1",
     "prettier-eslint": "^12.0.0",
+    "project-root-directory": "^1.0.3",
```
dustinbyrne, line 366: This is a 3 line module with ~10 weekly downloads. Maybe we consider just copying it into our project?

dustinbyrne (approved):

kgilpin:
That's what I usually do :-)

Issue #454, opened 2022-09-08
OpenAPI generation in VS Code returning invalid yaml


petecheslock:
This was previously confirmed working in `0.40.0` of the extension but there could have been a regression elsewhere? I'm able to recreate this error with the released `0.40.0` or later of the extension.  

<img width="473" alt="Screen Shot 2022-09-08 at 9 06 50 AM" src="https://user-images.githubusercontent.com/511733/189130726-21e558b7-9b13-463a-9c61-cee499c27bde.png">
[openapi-vscode-bad.yml.zip](https://github.com/applandinc/vscode-appland/files/9527049/openapi-vscode-bad.yml.zip)

Using https://github.com/land-of-apps/sample_app_6th_ed/ with `codespaces-devcontainer ` branch to recreate the problem.

PR #455, opened 2022-09-08
fix: Prevent a case where OpenAPI would generate invalid YAML

dustinbyrne:
If the document was large enough, line breaks may appear in unexpected places. This pull request fixes this issue. This change includes a test using the same data the bug was reported with and is now parsing the generated YAML to spot syntax errors.

image

Resolves https://github.com/applandinc/vscode-appland/issues/454


src/services/nodeDependencyProcess.ts
```diff
@@ -86,19 +86,19 @@ export class ProcessLog extends Array<ProcessLogItem> {
...
     }
+
     if (this.outputChannel) {
-      this.outputChannel.appendLine(`${this.process?.pid} [${OutputStream[stream]}] ${data}`);
+      const { outputChannel } = this;
+      data
+        .trimEnd()
+        .split('\n')
+        .forEach((line) => {
+          outputChannel.appendLine(`${this.process?.pid} [${OutputStream[stream]}] ${line}`);
```
kgilpin, line 100: How do you know that `line` is a complete line? It seems like `process.stdout.on('data', (data) => ` could emit partial lines, but those will be written to outputChannel as complete lines.

src/services/nodeDependencyProcess.ts
```diff
@@ -65,8 +65,8 @@ export class ProcessLog extends Array<ProcessLogItem> {
 
     process.stdout.setEncoding('utf8');
     process.stderr.setEncoding('utf8');
-    process.stdout.on('data', (data) => log.appendLines(data, OutputStream.Stdout, true));
-    process.stderr.on('data', (data) => log.appendLines(data, OutputStream.Stderr, true));
+    process.stdout.on('data', (data) => log.append(data, OutputStream.Stdout, true));
+    process.stderr.on('data', (data) => log.append(data, OutputStream.Stderr, true));
```
kgilpin, line 69: Does the third argument `cache` mean "save this data so that it can be written to the OpenAPI YAML file later?". If so, I don't think we want to save the STDERR output. The CLI command should be writing YAML only to stdout, and using stderr for things like error reporting.

kgilpin: I guess it's being filtered out here - https://github.com/applandinc/vscode-appland/blob/cb0054e5c03f70780017eaee3da4972bd1a62b1d/src/commands/generateOpenApi.ts#L84

src/services/nodeDependencyProcess.ts
```diff
@@ -86,19 +86,19 @@ export class ProcessLog extends Array<ProcessLogItem> {
...
 
-  protected appendLines(data: string, stream = OutputStream.Stdout, cache = false): void {
-    data
-      .trimEnd()
-      .split('\n')
-      .forEach((l) => this.append(l.trimEnd(), stream, cache));
-  }
-
   append(data: string, stream = OutputStream.Stdout, cache = false): void {
     if (this.shouldCache && cache) {
```
kgilpin, line 90: Maybe "cache" could be renamed to "saveOutput" or something like that? Or it could be a list of streams to save, like `[]`, `[ OutputStream.Stdout ]`, or something.

kgilpin (approved):

kgilpin:
The Travis job failed so I am restarting it.

src/services/nodeDependencyProcess.ts
dustinbyrne: It's possible `data` would be segmented and break up a line but this block is used for writing to the output tab and isn't parsed anything. There may be more than one process emitting logs to this window, so we're prefixing them with the process ID.

PR #465, opened 2022-09-20
fix: Robustly check the Instruction page titles

kgilpin:
Looks like there is a race condition on checking the Instructions page titles.

Also:

* Update `@appland` dependencies
* Filter out `uptodate` response lines which are blank after `trim`


apotterri (approved):
LGTM. Dunno if you also want the othet reviewers to respond, too....

PR #472, opened 2022-09-29
feat: Authenticate with AppMap Server

kgilpin:


test/integration/util.ts
```diff
@@ -126,12 +126,16 @@ export async function restoreFile(filePath: string, workspaceDir = ProjectA): Pr
...
  * Touches a file, then waits for appmap services to react.
+ *
  * @param touchFile file to touch
  * @returns current state of services
  */
 export async function waitForAppMapServices(touchFile: string): Promise<AppMapService> {
   const appMapService = await waitForExtension();
-  const wsPath = vscode.workspace.workspaceFolders![0].uri.fsPath;
+  assert(vscode.workspace.workspaceFolders, 'vscode.workspace.workspaceFolders');
+  assert(vscode.workspace.workspaceFolders[0], 'vscode.workspace.workspaceFolders[0]');
```
apotterri, line 136: Should this also be checking that `vscode.workspace.workspaceFolders.length === 1`? Or is it only important that there be at least one?

I'm also a little surprised that TS can infer from `assert(vscode.workspace.workspaceFolders` that `assert(vscode.workspace.workspaceFolders` is not `undefined`. I mean, of course it is, I just wouldn't have expected TS to be able to figure that out.

apotterri: Was there another TS warning here? What was it?

src/util.ts
```diff
@@ -23,6 +23,8 @@ export function getNonce(): string {
 // getStringRecords({ a: 'hello', b: [object Object] }, 'myApp') ->
 // { 'myApp.a': 'hello' }
 export function getRecords<T>(obj: Record<string, unknown>, keyPrefix?: string): Record<string, T> {
+  if (!obj) return {};
```
apotterri, line 26: It looks weird to have the type of `obj` not allow it to be `undefined`, but to have to do this check anyway.

How does this happen in practice? Maybe it could use a comment here, or in the caller, to explain what's going on?

apotterri (approved):
I had a couple of questions, but nothing of much substance. Looks ready to go.

src/util.ts
kgilpin: I’m not sure how it happens I just honestly saw it blowing up here and added this check.

test/integration/util.ts
kgilpin: Yes, TS does know that after an expression has been asserted, it’s not null. I use this kind of guard clause in various places.

PR #480, opened 2022-10-15
Improve formatting of findings tree view

### LINKED ISSUE #478: Runtime Analysis Formatting breaks when secret finding is multi-line
VS Code extension `0.46.0`
[Using the sample ruby on rails project](https://github.com/land-of-apps/sample_app_6th_ed/) the codespaces-devcontainer branch. 

When running tests `bundle exec rake test` there is a multi-line finding for secret in logs and it breaks the display on the left column view. 

Screen Shot 2022-10-13 at 4 38 29 PM
 
Screen Shot 2022-10-13 at 9 28 16 AM

dividedmind:
(Note https://github.com/getappmap/appmap-js/pull/788 is needed to fix the build.)

Before:
broken-findings-tree
After:
fixed-findings-tree

Fixes #478 

I also noticed another problem while working on this: #479

### DIFF
--- a/src/lib/firstLine.ts
+++ b/src/lib/firstLine.ts
@@ -1,0 +1,5 @@
+export default function firstLine(text: string | undefined): string | undefined {
+  const newline = text?.indexOf('\n');
+  if (newline === -1) return text;
+  return text?.substring(0, newline);
+}
--- a/src/lib/memoize.ts
+++ b/src/lib/memoize.ts
@@ -1,0 +1,15 @@
+/** Convert a function into a memoized form.
+ * WeakMap is used to keep the memoized inputs so as not to prevent GC.
+ */
+export default function // eslint-disable-next-line @typescript-eslint/ban-types
+memoize<T extends object, U>(fn: (x: T) => U): (x: T) => U {
+  const cache = new WeakMap<T, U>();
+  return (x) => {
+    let result = cache.get(x);
+    if (result === undefined) {
+      result = fn(x);
+      cache.set(x, result);
+    }
+    return result;
+  };
+}
--- a/src/lib/memoize.ts
+++ b/src/lib/memoize.ts
@@ -1,15 +1,10 @@
+import ValueCache from './ValueCache';
+
 /** Convert a function into a memoized form.
  * WeakMap is used to keep the memoized inputs so as not to prevent GC.
  */
 export default function // eslint-disable-next-line @typescript-eslint/ban-types
 memoize<T extends object, U>(fn: (x: T) => U): (x: T) => U {
-  const cache = new WeakMap<T, U>();
-  return (x) => {
-    let result = cache.get(x);
-    if (result === undefined) {
-      result = fn(x);
-      cache.set(x, result);
-    }
-    return result;
-  };
+  const cache = new ValueCache<T, U>();
+  return (x) => cache.getOrStore(x, () => fn(x));
 }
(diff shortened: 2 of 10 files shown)


dividedmind:
Node build is blocked by https://github.com/getappmap/appmap-js/pull/788

src/lib/memoize.ts
```diff
@@ -0,0 +1,15 @@
+/** Convert a function into a memoized form.
+ * WeakMap is used to keep the memoized inputs so as not to prevent GC.
+ */
+export default function // eslint-disable-next-line @typescript-eslint/ban-types
+memoize<T extends object, U>(fn: (x: T) => U): (x: T) => U {
```
ahtrotta, line 7: This is very nifty, but it seems like it might add unnecessary complexity. Are we having performance issues with this part of the code? If so, then it definitely makes sense to do this. But if not, then I would vote for a simpler solution until we know that there's a performance problem here.

src/lib/uniq.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Return a new array by removing duplicate values.
+ * If a transform is provided, the elements are compared on the results of that function.
+ */
+export default function uniq<T, U>(array: ReadonlyArray<T>, xform?: (x: T) => U): T[] {
+  if (xform) return [...new Map<U, T>(array.map((x) => [xform(x), x])).values()];
```
ahtrotta, line 6: Can you split this line up so that it's more readable? Also, why are you using a `Map` instead of a `Set`?
```
const xformedArray = array.map((x) => xform(x));
return [...new Set<T>(xformedArray)];
```

src/lib/firstLine.ts
```diff
@@ -0,0 +1,5 @@
+export default function firstLine(text: string | undefined): string | undefined {
+  const newline = text?.indexOf('\n');
```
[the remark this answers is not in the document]
dividedmind: We don't, but I'd be very surprised to see an `\r` in there. Regardless, it shouldn't really matter in this case; it still cuts the first line (it's harmless if it ends in whitespace).

src/lib/memoize.ts
dividedmind: Is it really that complex? Seems rather simple to me. The alternative is either recomputing it every time, which is rather icky (and note it is used as part of the sort key, so it's even more than one would think at the first glance) or sprinkling duplicated memoization code throughout (which is rather more unnecessary complexity).

It's not that we have performance issues here, but remember that it's JS, computing those strings creates a lot of unnecessary garbage. Plus it's really inelegant.

src/lib/uniq.ts
ahtrotta: Ah, right, I forgot we were just comparing based on the value returned by `xform` and not returning an array of the returned values. :+1:

src/lib/firstLine.ts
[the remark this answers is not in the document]
ahtrotta: :+1:

src/lib/memoize.ts
ahtrotta: I didn't mean that the functions were complex functions, just that it creates more misdirection when reading the code (moreso with the `ValueCache` than `memoize`). But I agree that that the individual functions aren't complex. 

I see your point, though, and they could be helpful for other parts of the code at some point.

ahtrotta (approved):

PR #482, opened 2022-10-17
Properly escape paths during installation

### LINKED ISSUE #481: Properly escape file paths when running the install command

ahtrotta:
Fixes #481

### DIFF
--- a/src/commands/installAgent.ts
+++ b/src/commands/installAgent.ts
@@ -8,20 +8,63 @@ export const InstallAgent = 'appmap.installAgent';
 const ELECTRON_COMMAND_PLATFORMS = ['linux', 'darwin'];
 let pendingTask: Thenable<unknown> | undefined;
 
-const escape = (str) => str.replace(/([^A-Za-z0-9_\-.,:/@\n])/g, '\\$1');
+type InstallEnv = {
+  ELECTRON_RUN_AS_NODE?: string;
+};
 
-function formatProjectPath(path: string): string {
-  if (path.includes(' ')) {
-    return "'" + path + "'";
+type InstallInformation = {
+  command: string;
+  env: InstallEnv;
+};
+
+function escapePath(str: string): string {
+  // on windows, quote everything except the drive letter when the path has one or more spaces
+  // e.g. C:"\Users\user\projects\directory with spaces"
+  if (os.platform() === 'win32') {
+    return str
+      .split(':')
+      .map((partialPath) => (partialPath.includes(' ') ? `"${partialPath}"` : partialPath))
+      .join(':');
   }
-  return path;
+
+  return str.replace(/([^A-Za-z0-9_\-.,:/@\n])/g, '\\$1');
 }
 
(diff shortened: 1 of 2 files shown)


test/integration/command/installAgent.test.ts
```diff
@@ -18,3 +21,188 @@ describe('Install agent command', () => {
...
+
+        const pythonCommand = generateInstallInfo(cwd, 'Python', false, globalStorageDir);
+        command = pythonCommand.command;
+        env = pythonCommand.env;
+        assert.deepStrictEqual(env, {});
+        assert.strictEqual(command, expected);
+      });
+    });
+
+    describe('without a space in the path', () => {
```
symwell, line 87: Should this class of testcases `without a space in the path` not have a space in the path for the globalStorageDir `C:"\\globalStorageDir\\with spaces in path"`?

test/integration/command/installAgent.test.ts
```diff
@@ -18,3 +21,188 @@ describe('Install agent command', () => {
...
+        const pythonCommand = generateInstallInfo(cwd, 'Python', false, globalStorageDir);
+        command = pythonCommand.command;
+        env = pythonCommand.env;
+        assert.deepStrictEqual(env, { ELECTRON_RUN_AS_NODE: 'true' });
+        assert(command.startsWith(expectedStart));
+        assert(command.includes(expectedEnd));
+      });
+    });
+
+    describe('without a space in the path', () => {
```
symwell, line 175: Same question here for spaces in the path for globalStorageDir for Unix.

symwell (approved):
Small questions.  Looks good to me.

src/commands/installAgent.ts
```diff
@@ -8,22 +8,65 @@ export const InstallAgent = 'appmap.installAgent';
...
+};
+
+function escapePath(str: string): string {
+  // on windows, quote everything except the drive letter when the path has one or more spaces
+  // e.g. C:"\Users\user\projects\directory with spaces"
+  if (os.platform() === 'win32') {
+    return str
+      .split(':')
+      .map((partialPath) => (partialPath.includes(' ') ? `"${partialPath}"` : partialPath))
+      .join(':');
```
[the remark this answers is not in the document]
ahtrotta: The split/map/join sequence is necessary because:
```
"C:\Users\user\projects\directory with spaces"
```
doesn't work, while:
```
C:"\Users\user\projects\directory with spaces"
```
does work.

test/integration/command/installAgent.test.ts
ahtrotta: I agree that the name is confusing. What I really meant was `without a space in the cwd`. 

I didn't want to overcomplicate the test cases by having separate `describe` blocks for spaces in the `globalStorageDir` in addition to what's already there, so I decided to just always have spaces in the `globalStorageDir` to make sure that we're handling that. But I think you're right and I'll change the `globalStorageDir` for that `describe` block.

PR #488, opened 2022-10-18
fix: Finding counts in the instructions view is consistent

### LINKED ISSUE #485: Findings and Problem counts are inconsistent in VS Code

dividedmind:
Fixes #485

### DIFF
--- a/src/services/findingsIndex.ts
+++ b/src/services/findingsIndex.ts
@@ -5,6 +5,7 @@ import { promisify } from 'util';
 import { readFile } from 'fs';
 import EventEmitter from 'events';
 import { fileExists } from '../util';
+import uniq from '../lib/uniq';
 
 export default class FindingsIndex extends EventEmitter implements vscode.Disposable {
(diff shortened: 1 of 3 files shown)


ahtrotta (approved):
This looks good to me. I'm surprised that `this.findingsByWorkspace.get(workspaceFolder)` yields an array with duplicate findings, though. Maybe we should open an issue for that?

dustinbyrne (changes requested):
This change should be rebased with master.

The `findingsByWorkspace` map was removed because it was problematic. Instead the findings are now retrieved dynamically by looking at the source URI in `findingsBySourceUri`.

dividedmind:
> This change should be rebased with master.
> 
> The `findingsByWorkspace` map was removed because it was problematic. Instead the findings are now retrieved dynamically by looking at the source URI in `findingsBySourceUri`.

Will do, thanks!

dustinbyrne (approved):

PR #491, opened 2022-10-19
fix: install button works with pyenv

### LINKED ISSUE #487: VS Code automated install fails when pyenv is in use
Loom [here](https://www.loom.com/share/69eb3cdbcb4041c19f3b5980d439a734).

When pyenv is in use, the automated installer on my M1 Mac auto-answers the first interactive console question with "No", which quits the installer.

ahtrotta:
Fixes #487


dustinbyrne (approved):
This seems somewhat problematic in that we're possibly preventing the user's environment from becoming active in the new terminal.

Maybe in the future we can activate the environment ourselves.

The Python extension itself seems to resort to `sleep`:
https://github.com/microsoft/vscode-python/blob/3698950c97982f31bb9dbfc19c4cd8308acda284/src/client/common/terminal/activator/base.ts#L41-L45

PR #493, opened 2022-10-20
fix: wait for terminal activation before sending install command

### LINKED ISSUE #494: Simpler solution for spawning terminals for pyenv users

ahtrotta:
Fixes #494

### DIFF
--- a/src/commands/installAgent.ts
+++ b/src/commands/installAgent.ts
@@ -6,19 +6,6 @@ import { NodeProcessService } from '../services/nodeProcessService';
 
 export const InstallAgent = 'appmap.installAgent';
 const ELECTRON_COMMAND_PLATFORMS = ['linux', 'darwin'];
-
-// these are in order of precedence
-const CONFIG_LOCATIONS = [
-  'workspaceFolderLanguageValue',
-  'workspaceLanguageValue',
-  'globalLanguageValue',
-  'defaultLanguageValue',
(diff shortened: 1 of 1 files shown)


src/commands/installAgent.ts
```diff
@@ -30,39 +17,15 @@ type InstallInformation = {
...
-function createTerminal(command: string, path: string, env: InstallEnv): void {
+async function createTerminal(command: string, path: string, env: InstallEnv): Promise<void> {
   const terminal = vscode.window.createTerminal({
     name: 'install-appmap',
     cwd: path,
     env,
   });
   terminal.show();
+  // wait for terminal activation, if any (e.g. `pyenv shell <VERSION>`)
+  await new Promise((resolve) => setTimeout(resolve, 2500));
```
dustinbyrne, line 28: Have you tested any lower values? I don't believe that we need to wait for the activation to occur, just for the commands to be sent so they don't get sent to the `stdin` of our own process.

ahtrotta: Anything less than `700` doesn't work on my machine, so I put `2500` just to be safe, but we could probably do lower. Maybe `1500` is better. What do you think?

dustinbyrne:
Let's wait and see what happens with the current release before resorting to sleep. It seemed like we only needed to wait for the activation command to be sent, but with having to wait this long it seems like there's more going on.

PR #501, opened 2022-10-27
fix: Strip paths of any extra leading path separators

dustinbyrne:
I'm not sure how this is possible, but some user's report an error similar to:
```
RangeError: path should be a `path.relative()`d string, but got "/yarn.lock"
```

This occurs regardless of the OS, sometimes using Windows path separators (e.g. `\yarn.lock`).

These paths must have double separators? I'm not sure how it's occurring, but this should fix it.

### DIFF
--- a/src/telemetry/properties/versionControlGit.ts
+++ b/src/telemetry/properties/versionControlGit.ts
@@ -30,7 +30,7 @@ export default class GitProperties implements VersionControlProperties {
     return !!this.ignores.find((ignore) => {
       if (!filePath.startsWith(ignore.dir)) return false;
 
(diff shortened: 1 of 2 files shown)


dividedmind (changes requested):
I'd really like to see a unit test here.

dustinbyrne:
Added. Any ideas on how this issue might occur given the line of code that was changed?

dividedmind:
> Added. 

Thanks, although I don't see any `//path` patterns in the test?

> Any ideas on how this issue might occur given the line of code that was changed?

My guess would be that a tool or a template adds `//yarn.lock` to gitignore.

PR #505, opened 2022-11-01
feat: Sequence diagram

kgilpin:
* Generate sequence diagram - with integration test
* Compare sequence diagrams - with integration test
* Copy AppMap to collection
* Delete individual AppMap
* Delete AppMap collection contents

PlantUML jar file and license are bundled in the extension. This should be OK per the terms of GPL and the PlantUML license, since PlantUML is running in a separate process. For example, `git` is also GPL.

## Demo

https://www.loom.com/share/18adba1c8262466aa3e7032d65b62046


src/authentication/appmapServerAuthenticationProvider.ts
```diff
@@ -98,7 +98,7 @@ export default class AppMapServerAuthenticationProvider implements vscode.Authen
     ];
 
     for (const [index, authnStrategy] of authnStrategies.entries()) {
-      await authnStrategy.prepareSignIn();
```
dustinbyrne, line 101: Without the `await` we should remove the option to return a promise from the interface as well
https://github.com/getappmap/vscode-appland/blob/master/src/authentication/authenticationStrategy/index.ts#L6

dustinbyrne (approved):

PR #525, opened 2022-11-11
Prompt users to install if the project is supported

dustinbyrne:
Users who open a project which is supported by AppMap, but contains no `appmap.yml` will receive a notification prompting them to install.

image

This popup will not open for new users immediately after installation as we're already opening the instructions automatically. Maybe it's worth revisiting this?

Clicking `Open instructions` will open the instructions view to the project picker.
Clicking `Later` will simply close the notification. It will appear again the next time the project is opened.
Clicking `Don't show again` will silence the notification, but only for the currently opened project.

### DIFF
--- a/src/actions/promptInstall.ts
+++ b/src/actions/promptInstall.ts
@@ -1,4 +1,4 @@
-import ProjectStateService from '../services/projectStateService';
+import ProjectStateService, { ProjectStateServiceInstance } from '../services/projectStateService';
 import { WorkspaceServices } from '../services/workspaceServices';
 import * as vscode from 'vscode';
 import { INSTALL_PROMPT, Telemetry } from '../telemetry';
@@ -22,6 +22,12 @@ const promptResponses: ReadonlyArray<vscode.MessageItem> = [
   { title: ButtonText.DontShowAgain },
 ];
 
+const meetsPromptCriteria = (project: ProjectStateServiceInstance): boolean =>
+  project.installable &&
+  !project.metadata.agentInstalled &&
+  project.metadata.language?.name === 'Ruby' &&
+  project.metadata.webFramework?.name === 'Rails';
+
 export default async function promptInstall(
   services: WorkspaceServices,
   extensionState: ExtensionState
@@ -35,9 +41,7 @@ export default async function promptInstall(
     undefined;
   if (silencePrompt) return;
 
-  const numInstallable = projectInstances.filter(
-    (project) => project.installable && !project.metadata.agentInstalled
-  ).length;
+  const numInstallable = projectInstances.filter(meetsPromptCriteria).length;
   if (!numInstallable) return;
 
   const numProjects = projectInstances.length;
--- a/src/actions/promptInstall.ts
+++ b/src/actions/promptInstall.ts
@@ -1,0 +1,71 @@
+import ProjectStateService from '../services/projectStateService';
+import { WorkspaceServices } from '../services/workspaceServices';
+import * as vscode from 'vscode';
+import { INSTALL_PROMPT, Telemetry } from '../telemetry';
+import ExtensionState from '../configuration/extensionState';
+
+export enum ButtonText {
+  Confirm = 'Open instructions',
+  Deny = 'Later',
+  DontShowAgain = "Don't show again",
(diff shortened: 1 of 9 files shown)


src/actions/promptInstall.ts
```diff
@@ -0,0 +1,71 @@
...
+  extensionState: ExtensionState
+): Promise<void> {
+  const projectService = services.getService(ProjectStateService);
+  if (!projectService) return;
+
+  const projectInstances = services.getServiceInstances(projectService);
+  const silencePrompt =
+    projectInstances.find(({ folder }) => extensionState.getHideInstallPrompt(folder)) !==
+    undefined;
+  if (silencePrompt) return;
```
dividedmind, line 36: Are we sure we want to skip the prompt even if only one of the projects is silenced?

src/actions/promptInstall.ts
```diff
@@ -0,0 +1,71 @@
...
+  const response = await vscode.window.showInformationMessage(msg.join(' '), ...promptResponses);
+  Telemetry.sendEvent(INSTALL_PROMPT, { result: getKey(response?.title || ButtonText.Deny) });
+
+  if (response?.title === ButtonText.Confirm) {
+    /*
+    This should execute `InstallGuideWebView.command` but requiring this file ends up importing an SVG. Our `tsc`
+    compiled tests have no way of reading this file, so it'll result in a panic at runtime. For now, reference the
+    command directly.
+    */
+    await vscode.commands.executeCommand('appmap.openInstallGuide', 'project-picker');
```
dividedmind, line 67: This is probably fine in any case.

src/actions/promptInstall.ts
```diff
@@ -0,0 +1,71 @@
...
+  const entry = Object.entries(ButtonText).find(([, v]) => v === val);
+  if (!entry) return 'Deny';
+  return entry[0] as keyof typeof ButtonText;
+}
+
+const promptResponses: ReadonlyArray<vscode.MessageItem> = [
+  { title: ButtonText.Confirm },
+  { title: ButtonText.Deny, isCloseAffordance: true },
+  { title: ButtonText.DontShowAgain },
+];
```
dividedmind, line 23: This is quite convoluted. It could probably be simplified a lot by doing this instead:
```typescript
const PromptResponses = [
  { key: 'confirm', title: 'Open instructions' },
  { key: 'deny', title: 'Later' },
  { key: 'dont-show', title: "Don't show again" },
] as const;
```

This should typecheck correctly on `showInformationMessage(msg, ...[...PromptResponses])` (weird, I know) and allow the key to be accessed directly as `response.key`. Response keys can be extracted if needed as
```typescript
export type Responses = typeof PromptResponses[number]['key'];
```

src/services/workspaceServices.ts
```diff
@@ -90,7 +90,8 @@ export class WorkspaceServices implements vscode.Disposable {
   getService<
     ServiceInstanceType extends WorkspaceServiceInstance,
     ServiceType extends WorkspaceService<ServiceInstanceType>
-  >(c: { new (): ServiceType }): ServiceType | undefined {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  >(c: { new (...args: any[]): ServiceType }): ServiceType | undefined {
```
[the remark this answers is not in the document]
dustinbyrne: That was my first impression as well and I did try to avoid the use of `any`. The compiler isn't keen on implicitly casting known argument types of the constructor to `unknown`. E.g.:
```log
TS2345: Argument of type 'typeof ProjectStateService' is not assignable to parameter of type 'new (...args: unknown[]) => ProjectStateService'.
  Types of construct signatures are incompatible.
    Type 'new (extensionState: ExtensionState, appMapWatcher: FileChangeEmitter, configWatcher: FileChangeEmitter, appMapCollection: AppMapCollection, classMapIndex: ClassMapIndex, findingsIndex?: FindingsIndex | undefined) => ProjectStateService' is not assignable to type 'new (...args: unknown[]) => ProjectStateService'.
      Types of parameters 'extensionState' and 'args' are incompatible.
        Type 'unknown' is not assignable to type '
```

PR #527, opened 2022-11-14
Unpin end-user dependencies (scanner, CLI)

dustinbyrne:
- `@appland/appmap` and `@appland/scanner` dependencies are now unpinned and will always be updated upon extension start
- A locking mechanism has been added such that only one VS Code process will attempt the installation/upgrade at a time. Other processes will wait until the lock has been removed before declaring dependencies ready. Tests have been added that exercise this process.


extern/yarn.js
```diff
@@ -1,4 +1,31 @@
...
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
dividedmind, line 28: Well caught!

src/services/nodeProcessService.ts
```diff
@@ -104,9 +111,48 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+      // Only let one process continue.
+      await lockfile.lock(this.globalStorageDir);
+    } catch (e) {
+      // We didn't aquire the lock, so we'll assume that another process is already installing.
+      // We'll just wait for it to finish.
+      log.appendLine('Another process is currently installing dependencies. Waiting...');
+
+      for (;;) {
+        try {
+          await lockfile.check(this.globalStorageDir);
```
dividedmind, line 134: This returns true or false, you need to check the value.

src/services/nodeProcessService.ts
```diff
@@ -104,9 +111,48 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+    }
+
+    try {
+      // If multiple VS Code windows are open, we may have multiple processes racing to install.
+      // Only let one process continue.
+      await lockfile.lock(this.globalStorageDir);
+    } catch (e) {
+      // We didn't aquire the lock, so we'll assume that another process is already installing.
+      // We'll just wait for it to finish.
+      log.appendLine('Another process is currently installing dependencies. Waiting...');
```
dividedmind, line 130: Is it intended to use `appendLine` here and `append` in the other message? Neither of the messages contains a trailing newline.

src/services/nodeProcessService.ts
```diff
@@ -164,6 +211,8 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
         exception: err,
         errorCode: ErrorCode.DependencyInstallFailure,
       });
+    } finally {
+      await lockfile.unlock(this.globalStorageDir);
```
dividedmind, line 215: You should probably use `release` instead, and only in the lock-acquired branch. With unlock you have a race – consider this sequence:
1. process alpha acquires the lockfile, starts installation,
2. beta sees the lock, waits around
3. alpha finishes, unlocks
4. beta notices the unlock and proceeds
5. gamma now acquires the lock and starts installation
6. beta reaches finally and releases the lock acquired by gamma
7. gamma is still installing despite no lock.

dividedmind (changes requested):
Nice work with the locking! Do you think this could be the cause of the EBUSY errors on Windows during the installation? I don't know if people often run several vscode processes. (If you open a new window, is that a new process?)

Needs some small changes and then it should be good to go! (Once the build is green.)

src/services/nodeProcessService.ts
```diff
@@ -104,9 +111,48 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
       await fs.mkdir(this.globalStorageDir, { recursive: true });
+    } catch {
+      // Ignore this error, it's probably because the directory already exists.
+      log.appendLine('Global storage directory already exists. Skipping.');
+    }
+
+    try {
+      // If multiple VS Code windows are open, we may have multiple processes racing to install.
+      // Only let one process continue.
+      await lockfile.lock(this.globalStorageDir);
```
dividedmind, line 126: BTW, you should probably hold onto the lock object. Otherwise it could get garbage collected and stop updating.

src/commands/generateOpenApi.ts
```diff
@@ -41,7 +42,19 @@ export default async function generateOpenApi(
...
           });
 
-          await verifyCommandOutput(openApiCmd);
+          try {
+            await verifyCommandOutput(openApiCmd);
+          } catch (e) {
+            Telemetry.sendEvent(DEBUG_EXCEPTION, {
+              exception: e as Error,
+              errorCode: ErrorCode.GenerateOpenApiFailure,
+              log: openApiCmd.log.toString(),
```
dividedmind, line 51: Is `openApiCmd.log` an array? If so, it might be a good idea to do `join('')` instead of `toString()`. I've seen those ugly commas if some of our telemetry logs, I think they might come from here.

dividedmind (changes requested):

src/commands/generateOpenApi.ts
dustinbyrne: It's an array, but has a custom `toString` method which joins it together similar to what you've suggested.

dividedmind (approved):
Thanks, looks good now!

PR #569, opened 2022-12-20
fix: Use `child_process.fork` for launching sub-processes

### LINKED ISSUE #546: Node sub-process fails to launch due to bad option `--ms-enable-electron-run-as-node`
e.g.

```
C:\Users\User\AppData\Local\Programs\Microsoft VS Code\Code.exe: bad option: --ms-enable-electron-run-as-node
```

```
C:\Program Files\Microsoft VS Code\Code.exe: bad option: --ms-enable-electron-run-as-node
```

```
/private/var/folders/3e/ax3xzq10xc61ckc1dke_rk9aids2l3ru0/T/AppTranslocation/623982B3-A93b-415B-B4BB-C1953D73AE0A/d/Visual Studio Code.app/Contents/Frameworks/Code Helper (Renderer).app/Contents/MacOS/Code Helper (Renderer): bad option: --ms-enable-electron-run-as-node
```

dustinbyrne:
This means we don't need to specify `--ms-enable-electron-run-as-node` or `ELECTRON_RUN_AS_NODE`. It also normalizes logic between remote environments.

Resolves #546

### DIFF
--- a/src/services/nodeDependencyProcess.ts
+++ b/src/services/nodeDependencyProcess.ts
@@ -16,9 +16,8 @@ export type WithLogCache = {
 export type ChildProcess = childProcess.ChildProcessWithoutNullStreams & WithLogCache;
 
 export type SpawnOptions = {
-  // Path to a bin script installed via yarn. If specified, it will be shifted into the command args
-  // to be run via `node`
-  binPath: string;
+  // Path to a node module script to be executed
+  modulePath: string;
 
   // Command line args given to `node` or the `bin` script specified
   args?: string[];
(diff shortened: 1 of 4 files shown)


dustinbyrne:
This needs to be tested on Mac and Windows

dustinbyrne:
Verified working on Mac, Windows and Linux.

ahtrotta (approved):

src/services/nodeDependencyProcess.ts
```diff
@@ -18,7 +18,7 @@ export type ChildProcess = childProcess.ChildProcessWithoutNullStreams & WithLog
 export type SpawnOptions = {
   // Path to a bin script installed via yarn. If specified, it will be shifted into the command args
```
dividedmind, line 19: This comment looks like it's out of date now: binPath is no longer optional. (BTW, perhaps `modulePath` would be a more accurate name?)

dividedmind (approved):

src/services/nodeDependencyProcess.ts
dustinbyrne: Great point, thanks! I've renamed `binPath` to `modulePath` across the board.

PR #573, opened 2023-01-04
fix: Override environment yarn configuration

### LINKED ISSUE #553: `Error: Failed to install dependencies... Command "up" not found.`

dustinbyrne:
Presumably some users have a yarn configuration which interferes with our dependency install. This change scrubs the environment of `YARN` configuration which may conflict.

Fixes #553. Integration tests are now run with `YARN_LOCKFILE_NAME` to make sure the override works. Stubbing `process.env` is not possible from the integration test.

### DIFF
--- a/src/services/nodeProcessService.ts
+++ b/src/services/nodeProcessService.ts
@@ -212,7 +212,13 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
         log: NodeProcessService.outputChannel,
         // Fix "The remote archive doesn't match the expected checksum" issue by
         // forcing yarn to fetch from the remote registry if checksums don't match.
-        env: { YARN_CHECKSUM_BEHAVIOR: 'update' },
+        env: {
+          YARN_CHECKSUM_BEHAVIOR: 'update',
+          YARN_ENABLE_TELEMETRY: 'false',
+          YARN_LOCKFILE_NAME: undefined,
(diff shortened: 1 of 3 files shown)


dividedmind (approved):

src/services/nodeProcessService.ts
```diff
@@ -212,7 +212,13 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
         log: NodeProcessService.outputChannel,
         // Fix "The remote archive doesn't match the expected checksum" issue by
         // forcing yarn to fetch from the remote registry if checksums don't match.
-        env: { YARN_CHECKSUM_BEHAVIOR: 'update' },
+        env: {
+          YARN_CHECKSUM_BEHAVIOR: 'update',
+          YARN_ENABLE_TELEMETRY: 'false',
+          YARN_LOCKFILE_NAME: undefined,
```
ahtrotta, line 218: Can we set this to the empty `yarn.lock` file generated above (line 206)? Alternatively, if we set `YARN_LOCKFILE_NAME: undefined`, do we need to generate the empty `yarn.lock` file above?

dustinbyrne: We set it to `undefined` to stop any value which may be present in `process.env` from propagating to `yarn up`. `yarn.lock` should be the default value anyhow.

PR #586, opened 2023-01-10
Add appmap to recommended extensions

### LINKED ISSUE #537: Add AppMap as a recommended extension to the installed project
When someone installs, we should create or modify the `.vscode/extensions.json` file in their project and add appmap to it as well. [From here](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions):

```
{
  "recommendations": ["appland.appmap"]
}
```

Anyone else opening up the project will then be notified that they can easily install the extension to get its benefits:

Image

ahtrotta:
Fixes #537

The first time that a user closes a map, prompt the user to add the AppMap extension to the list of recommended extensions for the current project in extensions.json in the .vscode folder.

image

If they click `Yes`, then the `extensions.json` file is opened in the editor:

image

### DIFF
--- a/src/appMapService.ts
+++ b/src/appMapService.ts
@@ -7,6 +7,7 @@ import ExtensionState from './configuration/extensionState';
 import AppMapEditorProvider from './editor/appmapEditorProvider';
 import { FindingsService } from './findingsService';
 import { AppMapConfigWatcher } from './services/appMapConfigWatcher';
+import { AppMapRecommenderService } from './services/appmapRecommenderService';
 import { AppmapUptodateService } from './services/appmapUptodateService';
 import Command from './services/command';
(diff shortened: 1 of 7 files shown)


dustinbyrne:
Modifying or creating `.vscode/extensions.json` on start up without informing or asking the user seems a little too aggressive IMO. Personally, I'd be likely to throw out any changes to this file without even looking at the diff if I saw it was modified. It doesn't happen often (if at all), and if it's changed, it's likely on accident.

This is purely speculation, but I think if we were to prompt the user after they've received value (closed their first AppMap?) we'd have better results.

kgilpin:
At least if the user discards the change, they know how to suggest this
extension to their team.

The work is done now, we must have done it for a reason?

dustinbyrne (approved):

PR #592, opened 2023-01-13
fix: Update @appland/appmap and @appland/scanner properly

### LINKED ISSUE #590: VS Code users appear not to be updating
End-user dependencies seem to have a long lag time between updates. This should not be the case. Updates (if available) should be fetched upon initialization.

The install/update process is defined here: https://github.com/getappmap/vscode-appland/blob/master/src/services/nodeProcessService.ts#L129

dustinbyrne:
Previously, `yarn up` was not giving the desired outcomes. Some users were not receiving updates. This command ensures both packages are updated while not affecting the version constraints defined in `package.json`.

To see this in action, check the version of `@appland/appmap` or `@appland/scanner` in `yarn.lock`.

On Linux:
```sh
$ cat ~/.config/Code/User/globalStorage/appland.appmap/yarn.lock | grep -A1 '@appland/appmap@npm:^3'
"@appland/appmap@npm:^3":
  version: 3.53.0
```

On Darwin:
```sh
$ cat ~/Library/ApplicationSupport/Code/User/globalStorage/appland.appmap/yarn.lock | grep -A1 '@appland/appmap@npm:^3'
"@appland/appmap@npm:^3":
  version: 3.53.0
```

Run the extension. The `AppMap: Services` tab in `Output` should indicate success installing dependencies. Checking the version again should yield an updated version:

```sh
$ cat ~/.config/Code/User/globalStorage/appland.appmap/yarn.lock | grep -A1 '@appland/appmap@npm:^3'
"@appland/appmap@npm:^3":
  version: 3.60.1
```

Fixes #590

### DIFF
--- a/src/services/nodeProcessService.ts
+++ b/src/services/nodeProcessService.ts
@@ -207,7 +207,7 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
 
       const installProcess = await spawn({
         modulePath: this.yarnPath,
-        args: ['up'],
+        args: ['up', '-R', '@appland/appmap', '@appland/scanner'],
         cwd: this.globalStorageDir,
         log: NodeProcessService.outputChannel,
         // Fix "The remote archive doesn't match the expected checksum" issue by


symwell (approved):

ahtrotta (approved):

src/services/nodeProcessService.ts
```diff
@@ -207,7 +207,7 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
 
       const installProcess = await spawn({
         modulePath: this.yarnPath,
-        args: ['up'],
+        args: ['up', '-R', '@appland/appmap', '@appland/scanner'],
```
apotterri, line 210: Per the doc:
```
If `-R,--recursive` is set the command will change behavior and no other switch 
will be allowed. When operating under this mode `yarn up` will force all ranges 
matching the selected packages to be resolved again (often to the highest 
available versions) before being stored in the lockfile. It however won't touch 
your manifests anymore, so depending on your needs you might want to run both 
`yarn up` and `yarn up -R` to cover all bases.
```

Do we need to run both `yarn up` and `yarn up -R`?

dustinbyrne: `yarn up` without the recursive flag would modify version constraints in `package.json`, I'm guessing that's what it's referring to by touching manifests?

In any case, I don't see any point to updating the version constraints. We've pinned them to major versions (`^1`, `^3`) and that seems fine. Ideally I don't want `package.json` changing, it should be fully controlled by the extension. It shouldn't matter if `^3` changes to `^3.53.0`, but I feel better knowing that every user has the same `package.json` - it's just one less variable to worry about.

symwell:
This might become a problem.

```
$ yarn --version
2.4.3
$ yarn up -R @appland/appmap @appland/scanner
Unknown Syntax Error: Unsupported option name ("-R").

$ yarn up [-i,--interactive] [-E,--exact] [-T,--tilde] [-C,--caret] ...
```

It worked with yarn 3.2.1.

PR #606, opened 2023-01-26
fix: Respect `appmap_dir` when generating OpenAPI

dustinbyrne:
When generating OpenAPI from within the extension, `--appmap-dir` was using the absolute path of the project instead of respecting the `appmap_dir` set within `appmap.yml` (if present).


kgilpin (approved):

src/commands/generateOpenApi.ts
```diff
@@ -36,9 +37,16 @@ export default async function generateOpenApi(
             dependency: ProgramName.Appmap,
             globalStoragePath: context.globalStorageUri.fsPath,
           });
+
+          let appmapDir = await lookupAppMapDir(workspaceFolder.uri.fsPath);
+          if (!appmapDir) {
+            appmapDir = '.';
```
kgilpin, line 43: Alternatively, if `appmap_dir` is missing, this argument can simply be omitted.

kgilpin (approved):

PR #619, opened 2023-02-15
fix: Use best fit path when opening source of a code object

### LINKED ISSUE #616: Can't open queries or requests
Opening an AppMap directly works, both in the Explore view and the sidebar.
But when clicking on an http request or sql query in Explore or the Code Objects view it tries to open a \.appmap.json or \\.appmap.json\ file respectively.
Clicking on one of the unittest methods inside the code tree in Code Objects also doesn't point to the right file. It just tries to append the package + java file path to the workspace.
Maybe it's because the main spring application and thus also the appmap.yml is contained in a subfolder of the workspace?

dustinbyrne:
Two main issues are fixed by this pull request:
1. When opening source code, the path was used verbatim. In Java, we only have partial paths (e.g. `myorg/myapp/package/Class.java`) so we should have been using `bestFilePath` when opening those files.
2. We were using URI fragments to send along the view state when opening an AppMap. This works everywhere but on Windows. The `URI` class will not allow us to create a properly formatted `file` URI containing a fragment on this platform. I'm not sure the usage is valid to begin with. Instead, a new command, `appmap.open` has been added to open an AppMap file with a particular state. We now use this instead.

I've added a frontend (system) test to guarantee we're getting visible selection of the intended item, however it seems a bug upstream is preventing us from setting the state properly. We'll need to address this in another pull request.  

Fixes #616

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -61,17 +61,6 @@ export default class AppMapEditorProvider
       })
     );
 
-    context.subscriptions.push(
-      vscode.commands.registerCommand('appmap.setAppmapStateNoPrompt', async (state) => {
-        if (provider.currentWebView) {
-          provider.currentWebView.webview.postMessage({
-            type: 'setAppmapState',
-            state: state,
-          });
-        }
-      })
-    );
-
     return provider;
   }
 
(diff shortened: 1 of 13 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -61,26 +61,15 @@ export default class AppMapEditorProvider
...
-    );
-
     return provider;
   }
 
   private static readonly viewType = 'appmap.views.appMapFile';
   private static readonly INSTRUCTIONS_VIEWED = 'APPMAP_INSTRUCTIONS_VIEWED';
   private static readonly RELEASE_KEY = 'APPMAP_RELEASE_KEY';
   private static readonly analysisManager = AnalysisManager;
+  private static readonly openWebviewPanels = new Map<string, vscode.WebviewPanel>();
```
ahtrotta, line 71: With the addition of `openWebviewPanels`, can we get rid of `documents`? It seems like it's superfluous now. I think you'd just have to change the implementation of the `openDocuments` getter method.

ahtrotta (changes requested):
Looks good! Just a couple of comments/questions.

src/editor/appmapEditorProvider.ts
dustinbyrne: AFAIK I cannot get the document from an instance of `WebviewPanel`. Unless you're suggesting the key be of type `AppMapDocument`, but that makes the interface difficult to use as the `AppMapDocument` class is coupled tightly with AppMap data.

ahtrotta (approved):

PR #636, opened 2023-02-24
fix: do not let users bypass sign-in

### LINKED ISSUE #637: Some users can create appmaps without signing in
A new user who has not yet signed in opening a new project will get an alert prompting them to install AppMap. From the alert they can navigate to the Project Picker and then next/back through the whole setup flow without signing in.

Image

ahtrotta:
Fixes #637


dustinbyrne (approved):

kgilpin:
If only we had AppMaps of these fixes.

PR #638, opened 2023-02-27
fix: reduce scanning for appmap.yml

### LINKED ISSUE #633: High CPU utilization running the VS Code extension
Every second, the extension scans the entire directory tree under the project root, looking for any files named `appmap.yml`. If the tree is big (e.g. `appmap-js`, with a fully-populated `node_modules`), this will cause very high CPU usage.

apotterri:
**This shouldn't be merged until the `wip` commit gets removed. It's only there as a temporary measure to run the integration tests.**

Reduce the cases where the extension scans the entire subtree under the current project directory.

Fixes #633 .

### DIFF
--- a/src/services/nodeProcessService.ts
+++ b/src/services/nodeProcessService.ts
@@ -1,25 +1,50 @@
-import * as vscode from 'vscode';
-import * as path from 'path';
-import os from 'os';
-import { DEBUG_EXCEPTION, Telemetry } from '../telemetry';
 import { createWriteStream, promises as fs } from 'fs';
-import { WorkspaceService } from './workspaceService';
-import NodeProcessServiceInstance from './nodeProcessServiceInstance';
-import { ProcessWatcher } from './processWatcher';
-import ErrorCode from '../telemetry/definitions/errorCodes';
-import { getModulePath, ProgramName, spawn } from './nodeDependencyProcess';
-import { downloadFile, getLatestVersionInfo } from '../util';
-import { lookupAppMapDir } from '../lib/appmapDir';
+import os from 'os';
+import * as path from 'path';
 import lockfile from 'proper-lockfile';
-import IndexProcessWatcher from './indexProcessWatcher';
-import ScanProcessWatcher from './scanProcessWatcher';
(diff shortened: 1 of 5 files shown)


src/services/nodeProcessService.ts
```diff
@@ -84,22 +84,42 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
       }
     };
 
+    const excludes = vscode.workspace.getConfiguration('files').get('watcherExclude') as string[];
```
dividedmind, line 87: Is it really a `string[]`? You seem to be using it with `Object.keys` later, which would give consecutive integers if that was the case.

dividedmind (changes requested):
There is some type weirdness happening there.

Also, I wonder if it wouldn't be a better design to move this logic into the `ConfigFileProvider` instead of littering the watchers with yet another parameter? Feels like it belongs there.

apotterri:
> There is some type weirdness happening there.
> 
> Also, I wonder if it wouldn't be a better design to move this logic into the `ConfigFileProvider` instead of littering the watchers with yet another parameter? Feels like it belongs there.

I didn't put it in `ConfigFileProvider`, because `ConfigFileProvider` is so simple:
```
export type ConfigFileProvider = () => Promise<vscode.Uri[]>;
```
It seems reasonable to expand it with this functionality, though. Let me try it and see what it looks like.

dividedmind:
> > There is some type weirdness happening there.
> > Also, I wonder if it wouldn't be a better design to move this logic into the `ConfigFileProvider` instead of littering the watchers with yet another parameter? Feels like it belongs there.
> 
> I didn't put it in `ConfigFileProvider`, because `ConfigFileProvider` is so simple:
> 
> ```
> export type ConfigFileProvider = () => Promise<vscode.Uri[]>;
> ```
> 
> It seems reasonable to expand it with this functionality, though. Let me try it and see what it looks like.

Thanks! You can probably keep the simple API (at least consumer-side), just put more logic there :)

dividedmind (approved):

PR #641, opened 2023-03-01
Fix sign-in webview rendering

### LINKED ISSUE #640: Sign-in webview is properly rendered
Before fix:

image

After fix:

image

ahtrotta:
Fixes #640 

After fix:

image


src/webviews/getWebviewContent.ts
```diff
@@ -12,7 +12,8 @@ export default function getWebviewContent(
   webview: vscode.Webview,
   context: vscode.ExtensionContext,
   title: string,
-  appmapModule: AppmapModule
+  appmapModule: AppmapModule,
+  htmlStyle?: string
```
dustinbyrne, line 16: Maybe this should be `htmlStyle = ''` otherwise if this value isn't provided you'd end up with:
```
<html style="undefined" lang="en">
```

Unless VS Code is re-writing this anyway?

ahtrotta: It looks like whatever gets defined there gets added to what VS Code writes. 

So here's the sidebar sign-in webview:

image

And here's a findings info page:

image

So it seems like the `undefined` value doesn't do anything. I can still change it if you want, though.

PR #655, opened 2023-03-22
Stats panel in appmap

### LINKED ISSUE #654: Stats panel in appmaps

ahtrotta:
Fixes #654


src/editor/AppMapDocument.ts
```diff
@@ -1,13 +1,23 @@
...
-  constructor(public uri: vscode.Uri, public raw: Uint8Array, public findings?: FindingInfo[]) {
+  constructor(
+    public uri: vscode.Uri,
+    public raw: Uint8Array,
+    stats: MapStats,
+    public findings?: FindingInfo[]
+  ) {
     const appMap = JSON.parse(raw.toString());
     if (findings && findings.length !== 0) appMap.findings = findings;
+    appMap.stats = stats || {};
```
dustinbyrne, line 20: This looks like a detail I missed in the frontend. Is there a reason `stats` is appended to the AppMap data instead of passed as it's own prop? In this fashion we can only ever generate the stats synchronously, i.e. they must be provided at the time we open the AppMap. Ideally we don't need to wait for the CLI to exit before opening an AppMap.

ahtrotta: Since we want stats for all maps, including giant ones that we won't load into memory, I wanted to generate stats in the `openCustomEditor` function so that they can be generated before trying to read the file.

I can change it so that the stats are generated after the file is opened, though. I could generate the stats and then send a message to the webview to update it's state. Does that sound good?

dustinbyrne: Let's see how it goes. The additional time may be a non-issue. If it turns out we need to do this asynchronously in the future, we can always just `stat` the file to check the size beforehand and act accordingly.

dustinbyrne (approved):

PR #656, opened 2023-03-29
Handle large appmaps

### LINKED ISSUE #651: Better handling for large and giant appmaps

ahtrotta:
Fixes #651

- [x] When opening an appmap > 10 MB, the extension will automatically prune the map down to 10 MB in memory, and then display the pruned map with notifications.

image

- [x] When opening an appmap > 200 MB, the extension will generate stats about the map, and then display an empty map opened to the Stats panel with notifications.

image

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -96,6 +96,7 @@ export default class AppMapEditorProvider
   public static readonly SEQ_DIAGRAM_FEEDBACK_REQUESTED = 'SEQ_DIAGRAM_FEEDBACK_REQUESTED';
   private static readonly LARGE_APPMAP_SIZE = 10 * 1000 * 1000; // 10 MB
   private static readonly GIANT_APPMAP_SIZE = 200 * 1000 * 1000; // 200 MB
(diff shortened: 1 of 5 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -102,25 +106,67 @@ export default class AppMapEditorProvider
...
     const stats = { functions };
 
+    // If the map is Giant, don't read it into memory
+    if (appmapFileSize > AppMapEditorProvider.GIANT_APPMAP_SIZE)
+      return new AppMapDocument(uri, '{}', stats, []);
+
+    let data = (await vscode.workspace.fs.readFile(uri)).toString();
+
+    // If the map is Large, automatically prune it to ~ 10 MB
+    if (appmapFileSize > AppMapEditorProvider.LARGE_APPMAP_SIZE) {
```
dustinbyrne, line 121: If the data needs to be pruned, we're about to throw out the results of this read file anyway. Maybe this should be an `if`/`else` or otherwise reverse order?

ahtrotta: Yeah, good point, I just pushed a fix.

dustinbyrne (approved):

PR #665, opened 2023-04-10
Improved support for monorepos

### LINKED ISSUE #664: VS Code extension should handle monorepos
Currently, we do not handle monorepos well, but we need to to be able to use our product in `appmap-js`. 

The plan is to run an instance of the scanner and indexer for each `appmap.yml` in the monorepo. Then we can scan multiple `appmap-dir`s within a monorepo and we can avoid scanning the entire project constantly.

ahtrotta:
Fixes #664 

With this change, the extension can handle monorepos with multiple `appmap.yml` files (like our own `appmap-js`).

- [x] Launch an instance of indexer and scanner for each `appmap.yml` file in the workspace.
- [x] Require the user to pick which folder to use for the OpenAPI button.
- [x] Require the user to pick which folder to use for the `Save To Collection` map action.
- [x] Re-launch indexer and scanner when an `appmap.yml` file is created, deleted, or changed.
- [x] Change the default location to execute indexer and scanner from `.` to `tmp/appmap`.


src/telemetry/definitions/properties.ts
```diff
@@ -105,8 +110,16 @@ export const FILE_METADATA = new TelemetryDataProvider({
 
 export const AGENT_CONFIG_PRESENT = new TelemetryDataProvider({
   id: 'appmap.agent.config_present',
-  async value({ rootDirectory }: { rootDirectory: PathLike }) {
-    const isPresent = await fileExists(path.join(rootDirectory.toString(), 'appmap.yml'));
+  async value({ uri }: { uri: vscode.Uri }) {
+    const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri);
+    if (!workspaceFolder) return String('false');
```
dustinbyrne, line 115: The string -> string cast isn't necessary
```ts
if (!workspaceFolder) return 'false';
```

src/telemetry/definitions/properties.ts
```diff
@@ -105,8 +110,16 @@ export const FILE_METADATA = new TelemetryDataProvider({
...
+
+    const configManager = workspaceServices().getServiceInstanceFromClass(
+      AppmapConfigManager,
+      workspaceFolder
+    ) as AppmapConfigManagerInstance | undefined;
+
+    const isPresent = configManager && !configManager.isUsingDefaultConfig;
```
dustinbyrne: If `isUsingDefaultConfig` is true, isn't it also true that they must have had an `appmap.yml`, it just didn't contain an `appmap_dir`?

If so, we should still report the config as present regardless of whether or not `appmap_dir` is set.

src/services/projectStateService.ts
```diff
@@ -221,8 +221,11 @@ export class ProjectStateServiceInstance implements WorkspaceServiceInstance {
   }
 
   private async configurationInWorkspace(): Promise<boolean> {
-    const existingConfigs = await promisify(glob)(`${this.folder.uri.fsPath}/**/appmap.yml`);
-    return existingConfigs.length !== 0;
+    const configManager = workspaceServices().getServiceInstanceFromClass(
+      AppmapConfigManager,
+      this.folder
+    ) as AppmapConfigManagerInstance | undefined;
+    return !!(configManager && !configManager.isUsingDefaultConfig);
```
dustinbyrne: Similar comment as what I wrote for the `appmap.agent.config_present` property - this should return true if `appmap.yml` is present, regardless of whether or not `appmap_dir` is set.

src/lib/saveAppMapToCollection.ts
```diff
@@ -22,21 +22,23 @@ export default async function saveAppMapToCollection(
...
 
-  if (appmapConfig.appmapDir && appmapConfig.configFolder) {
+  if (appmapConfig) {
     const appmapDirFullPath = join(appmapConfig.configFolder, appmapConfig.appmapDir);
     appmapDir = relative(projectFolder.uri.fsPath, appmapDirFullPath);
+    configFolder = appmapConfig.configFolder;
   }
 
-  if (!appmapDir || appmapConfigManagerInstance.isUsingDefaultConfig) {
+  if (appmapConfig?.usingDefault || appmapConfigManagerInstance.isUsingDefaultConfig) {
```
dustinbyrne, line 41: What's the difference?

src/commands/generateOpenApi.ts
```diff
@@ -50,9 +50,8 @@ export default async function generateOpenApi(
           assert(appmapConfigManagerInstance);
 
           const appmapConfig = await appmapConfigManagerInstance.getAppmapConfig();
-          if (!appmapConfig) return;
 
-          if (appmapConfig.appmapDir && appmapConfig.configFolder) {
+          if (appmapConfig && appmapConfig.appmapDir && appmapConfig.configFolder) {
```
dustinbyrne, line 54: It seems like these cases where `appmapDir` and `configFolder` are empty should be handled by the `AppMapConfigManager`. The fields aren't `null` or `undefined`, so if we're worried about empty values here, shouldn't they have some defaults set?

src/commands/generateOpenApi.ts
```diff
@@ -42,23 +42,13 @@ export default async function generateOpenApi(
...
-
-            configToUse = configs.find((config) => config.configFolder === pick);
-          }
-
-          if (configToUse && configToUse.appmapDir && configToUse.configFolder) {
-            appmapDir = configToUse.appmapDir;
-            cwd = configToUse.configFolder;
+          if (appmapConfig && appmapConfig.appmapDir && appmapConfig.configFolder) {
+            appmapDir = appmapConfig.appmapDir;
+            cwd = appmapConfig.configFolder;
```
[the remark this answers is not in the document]
ahtrotta: Yeah, I got a little sloppy with my git hygiene in this PR, sorry about that.

src/services/appmapConfigManager.ts
```diff
@@ -135,109 +147,88 @@ export class AppmapConfigManager {
...
-      const result = {} as AppmapConfig;
-
-      try {
-        const appmapConfig = load(await readFile(configFilePath, 'utf-8'));
-        assert(appmapConfig && typeof appmapConfig === 'object');
-
-        result.configFolder = dirname(configFilePath);
+  dispose(): void {
+    return;
+  }
```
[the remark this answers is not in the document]
ahtrotta: Yeah, I fixed this in a later commit.

src/services/appmapConfigManager.ts
```diff
@@ -135,109 +147,88 @@ export class AppmapConfigManager {
...
+      })
+    );
   }
 
-  private static setupWatcher() {
+  private setupWatcher(): void {
     if (!this._configWatcher || this._watcherConfigured) return;
 
-    this._configWatcher.onCreate(async ({ workspaceFolder, uri }) => {
+    this._configWatcher?.onCreate(async ({ workspaceFolder, uri }) => {
```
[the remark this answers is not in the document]
ahtrotta: I fixed this in a later commit.

src/lib/saveAppMapToCollection.ts
```diff
@@ -18,18 +18,26 @@ export default async function saveAppMapToCollection(
     vscode.window.showErrorMessage(`Cannot determine workspace folder for ${appmapUri.fsPath}`);
     return;
   }
-  let appmapDir: string | undefined = await lookupAppMapDir(projectFolder.uri.fsPath);
-  if (!appmapDir) {
+
+  let appmapDir: string | undefined;
+  const appmapConfig = await AppmapConfigManager.getAppmapConfigforWorkspace(projectFolder);
+
+  if (appmapConfig && appmapConfig.appmapDir && appmapConfig.configFolder) {
```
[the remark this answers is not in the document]
ahtrotta: Yeah, I went through and removed these unnecessary checks, thanks!

src/services/appmapConfigManager.ts
```diff
@@ -135,109 +147,88 @@ export class AppmapConfigManager {
...
   }
 
-  private static async configForWorkspace(
-    folder: vscode.WorkspaceFolder
-  ): Promise<WorkspaceConfig> {
-    const configPattern = new vscode.RelativePattern(folder, AppmapConfigManager.CONFIG_PATTERN);
-    const configFileProvider = new ConfigFileProviderImpl(configPattern);
-    const configFiles = await configFileProvider.files();
+  private async appMapConfigFromFile(configFilePath: string): Promise<AppmapConfig | undefined> {
+    if (await fileExists(configFilePath)) {
```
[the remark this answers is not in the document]
ahtrotta: Fixed!

src/services/nodeProcessService.ts
```diff
@@ -52,6 +51,39 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+  private async handleConfigChange(folder: vscode.WorkspaceFolder): Promise<void> {
+    const currentInstance = workspaceServices().getServiceInstanceFromClass(
+      NodeProcessService,
+      folder
+    ) as NodeProcessServiceInstance | undefined;
+    assert(currentInstance);
+
+    workspaceServices().unenrollServiceInstance(folder, currentInstance);
+    await currentInstance.stop();
+    currentInstance.dispose();
```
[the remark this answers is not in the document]
ahtrotta: Yup, good call

src/telemetry/definitions/properties.ts
ahtrotta: :upside_down_face:

src/services/appmapConfigManager.ts
```diff
@@ -135,109 +147,88 @@ export class AppmapConfigManager {
...
+      const result = {} as AppmapConfig;
```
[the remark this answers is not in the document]
ahtrotta: I changed it so the default values are present when `result` is defined, and if the config file has an `appmap_dir` then I overwrite the default value.

src/telemetry/definitions/properties.ts
ahtrotta: I changed this in a later commit to use `hasConfigFile`, which is more straightforward, I think.

src/services/projectStateService.ts
ahtrotta: I also changed this in a later commit to use `hasConfigFile` instead.

src/commands/generateOpenApi.ts
ahtrotta: Yeah, it is handled in the `AppmapConfigManager` so I removed these unnecessary checks.

src/lib/saveAppMapToCollection.ts
ahtrotta: The class property `isUsingDefaultConfig` is `true` when there is one or more `appmap.yml` files, but all of them return `undefined` from `appMapConfigFromFile` (if they're unreadable, for example). In that case, the config for the project is the default config.

The `usingDefault` property for the `AppmapConfig` object type is `true` if the `appmap.yml` is present and readable but does not have an `appmap_dir` specified.

src/services/appmapConfigManager.ts
```diff
@@ -134,6 +134,23 @@ export class AppmapConfigManager {
...
+      let appmapConfig: unknown;
+      try {
+        appmapConfig = load(await readFile(appmapConfigFilePath, 'utf-8'));
+        assert(appmapConfig && typeof appmapConfig === 'object');
+      } catch (e) {
+        // Unparseable AppMap config, or related error.
+        console.warn(e);
+        return;
+      }
+      await appendFile(appmapConfigFilePath, `appmap_dir: ${appmapDir}`);
```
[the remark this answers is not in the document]
ahtrotta: This method is only called `if (appmapConfig?.usingDefault || appmapConfigManagerInstance.isUsingDefaultConfig)` [here](https://github.com/getappmap/vscode-appland/pull/665/files#diff-6c54be8b9c319e99e28a8317315f5f901b5a3a2cc0d6a455391ea0a541d8e231R41), which means that there is not an `appmap_dir` already present in the file, because if there was an `appmap_dir` then we wouldn't be using a default value.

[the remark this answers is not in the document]
ahtrotta: I decided to merge the new `appmap_dir` into the existing config, as you said, just to be safe.

dustinbyrne (approved):

PR #682, opened 2023-05-10
fix: Use a stable node bin path when running the install CLI

dustinbyrne:
`process.argv0` could sometimes point towards an incorrect binary path
`process.execPath` should be more accurate, and it's now testing for presence of `code` in the command

### DIFF
--- a/src/commands/installAgent.ts
+++ b/src/commands/installAgent.ts
@@ -89,7 +89,9 @@ function escapePath(str: string): string {
 }
 
 function electronCommand(globalStorageDir: string, installLocation: string): string {
-  const nodePath = escapePath(process.argv0);
+  // `child_process.fork` uses `process.execPath` from the parent process to spawn a new process.
+  // This is the exact behavior we want to emulate in the terminal, so we'll use it here as well.
+  const nodePath = escapePath(process.execPath);
   const cliPath = join(globalStorageDir, 'node_modules', '@appland', 'appmap', 'built', 'cli.js');
   const flags = ['--ms-enable-electron-run-as-node', '-d', installLocation];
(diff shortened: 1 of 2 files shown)


kgilpin:
This doesn’t affect the index or scanner commands?

We didn’t get error reports through telemetry?

Is additional telemetry needed to alert us on this going forward?

ahtrotta:
> This doesn’t affect the index or scanner commands?
> 
> We didn’t get error reports through telemetry?
> 
> Is additional telemetry needed to alert us on this going forward?

The indexer and scanner use [`child_process.spawn`](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options), so they're not affected.

The exit code from the terminal isn't captured, so we didn't get error telemetry based on the outcome of the terminal process. We could wait for the terminal process to exit [here](https://github.com/getappmap/vscode-appland/blob/master/src/commands/installAgent.ts#L77) and then check the exit code and send telemetry for that. Dustin what do you think?

kgilpin:
I think on this matter we need all the telemetry we can get - even if it's
not 100% reliable (well, telemetry never is anyway)

kgilpin:
It's good to solve this problem but we also have to explain how we are
going to prevent it from recurring.

dustinbyrne:
They should be using `child_process.fork`, it's more resilient in this context because it's purpose is for spawning node modules.

I don't think we have the ability to capture the exit code of the process run in the terminal, just the terminal itself. Once the process completes (or fails) the terminal remains open. It doesn't propagate the exit code.

ahtrotta (approved):
My vote is to merge this fix and then look into improving telemetry afterwards.

PR #683, opened 2023-05-10
User can configure default appmap filter settings

### LINKED ISSUE #681: User can configure the default appmap filter

ahtrotta:
Fixes #681 

This incorporates the frontend changes from [this PR](https://github.com/getappmap/appmap-js/pull/1186) and adds logic to store and update saved filters.

image

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -40,6 +40,12 @@ export type FunctionStats = {
   location: string;
 };
 
+export type SavedFilter = {
+  filterName: string;
+  state: string;
+  default: boolean;
+};
+
 /**
  * Provider for AppLand scenario files.
  */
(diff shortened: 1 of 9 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -210,6 +217,56 @@ export default class AppMapEditorProvider
...
+
+  async defaultFilter(filter: SavedFilter): Promise<void> {
+    const savedFilters = this.getSavedFilters();
+    if (!savedFilters) return;
+
+    const newSavedFilters = savedFilters.map((savedFilter) => {
+      const defaultValue = savedFilter.filterName === filter.filterName;
+      savedFilter.default = defaultValue;
+      return savedFilter;
+    });
```
dustinbyrne, line 255: Nit - this should be a `forEach` as it's not transforming the value returned, just mutating it. Generally `map` shouldn't result in side effects.

test/integration/savedFilters.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    editorProvider = new AppMapEditorProvider(context, extensionState);
+    updateFiltersSpy = sinon.spy(editorProvider, 'updateFilters');
+    initializeDefaultFilter(context);
+  });
+  afterEach(() => sandbox.restore());
+
+  it('initializes filters to AppMap default', () => {
+    assert.deepEqual(context.workspaceState.get(AppMapEditorProvider.SAVED_FILTERS), [
+      defaultFilter,
+    ]);
```
dustinbyrne, line 41: I think this only passes because of the implementation of our mock `Momento` class:
https://github.com/getappmap/vscode-appland/blob/master/test/mocks/inMemoryMemento.ts#L12-L15

`initializeDefaultFilter` doesn't `await` the update or return a `Promise`. If it were actually hitting the disk I'd imagine this would fail.

dustinbyrne (changes requested):
Looks good. I'd really only like to see `initializeDefaultFilter` return a `Promise` to avoid any future issues.

src/editor/appmapEditorProvider.ts
ahtrotta: Good catch! I seem to have a bad habit of doing this :upside_down_face:

test/integration/savedFilters.test.ts
ahtrotta: Makes, sense, I changed it :+1:

PR #690, opened 2023-05-18
Add installability details to telemetry

### LINKED ISSUE #689: Telemetry includes information on why a project is not installable
When a project is not installable, it would greatly help to understand why. There are plenty of examples of projects that were not installable but were a supported language, for instance.

ahtrotta:
Fixes #689 

Adds information about why a project is or is not installable. Specifically, it adds details about how we judge web and test frameworks for a project.

### DIFF
--- a/src/telemetry/definitions/properties.ts
+++ b/src/telemetry/definitions/properties.ts
@@ -153,6 +153,20 @@ export const PROJECT_LANGUAGE = new TelemetryDataProvider({
   },
 });
 
+export const WEB_FRAMEWORK = new TelemetryDataProvider({
+  id: 'appmap.project.web_framework',
+  async value({ project }: { project: ProjectMetadata }) {
+    return project?.webFramework?.name;
+  },
+});
+
+export const TEST_FRAMEWORK = new TelemetryDataProvider({
+  id: 'appmap.project.test_framework',
+  async value({ project }: { project: ProjectMetadata }) {
+    return project?.testFramework?.name;
+  },
+});
+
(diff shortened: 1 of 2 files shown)


src/telemetry/definitions/properties.ts
```diff
@@ -205,6 +205,15 @@ export const IS_INSTALLABLE = new TelemetryDataProvider({
...
 
+export const INSTALLABILITY_DATA = new TelemetryDataProvider({
+  id: 'appmap.project.installability_data',
+  async value({ project }: { project: ProjectMetadata }) {
+    const { webFramework, testFramework, language } = project;
+    const result = { webFramework, testFramework, language };
+    return JSON.stringify(result);
+  },
+});
+
```
dustinbyrne, line 216: Can you split this out into two separate properties?
`appmap.project.web_framework` and `appmap.project.test_framework` (`appmap.project.language` already exists)

dustinbyrne (changes requested):

src/telemetry/definitions/properties.ts
dustinbyrne: It's easier to write queries for - otherwise we need to expand the object into new columns

ahtrotta: Yeah, that makes sense. Will do!

ahtrotta: Do you just want the name of the web framework as the value of `appmap.project.web_framework`?

dustinbyrne (approved):

PR #691, opened 2023-05-19
fix: Look for appmaps in subprojects

dividedmind:
A recent change limited places in which VSCode looks for AppMaps to hardcoded `/{tmp,target,build}/appmap`. This caused it to no longer pick up AppMaps in subprojects of a monorepo.

Now it looks for those directories in the full tree.


dividedmind:
> Monorepos should have an AppMap.yml file in each subproject. A single AppMap.yml at the root of the project is not a supported configuration.

This is beside the point. The code as it stands now looks for appmaps only in the root, regardless of whether there are any `appmap.yml`s anywhere.

kgilpin (approved):

PR #695, opened 2023-05-23
feat: Propagate VS Code identity to CLI

dustinbyrne:
This change specifies user id / session id overrides in the environment of background processes to ensure a single thread of user activity.

As it turns out, we don't need to override VS Code's implementation of Session ID. It doesn't appear to ever renew within the lifetime of the window.


dustinbyrne:
I'm going to move this back to a draft for now. Another requirement has become clear and it may affect this PR.

dividedmind (approved):

PR #705, opened 2023-06-05
feat: Broaden install prompt

kgilpin:
Prompt to install if a project satisfies all of:

* Has a supported language
* Has a supported framework
* Does not already have an appmap.yml

@dividedmind FYI


dividedmind:
Looks ok to me. BTW, you should probably pull https://github.com/getappmap/vscode-appland/issues/698 lest someone else will start doing it, too.

kgilpin:
@dustinbyrne let's check the telemetry data 7 days from now (June 15) and see if we should enable this for Python.

PR #707, opened 2023-06-05
feat: AppMap installer for Python

### LINKED ISSUE #702: Install using the Python terminal provided by VS Code if it is configured
Use terminal provided by “Python: Create Terminal”, provided by https://github.com/Microsoft/vscode-python. 

If not available, perhaps prompt the user to perform a manual install?

dustinbyrne:
The `Automated install via AppMap CLI` button now uses the Python extension to create a Python terminal to run the command in. The Python terminal will use the users own configured interpreter/virtual environment.

The overall flow is as follows:
image

If the user does not have the extension installed, they are prompted to install it.
image

Clicking `Yes` to install waits for the extension to install then continues on to the CLI installation. If the extension was already present, this step is skipped.
image

Before running the CLI, we check to see if the Python extension knows of any virtual environments within the project directory. If the user doesn't have one selected, we'll prompt them to reconsider:
image

The CLI installer is then run within a Python context:
image

Resolves #702


src/commands/installer/python.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    if (!this.pythonExtensionLoader) {
+      const result = await vscode.window.showInformationMessage(
+        'The AppMap extension requires the Python extension to be installed. Would you like to install it now?',
+        'Yes',
+        'No'
+      );
+
+      if (result === 'Yes') {
+        await this.installPythonExtension();
```
kgilpin, line 120: I have the extension, but it is disabled for this workspace. That seems to break this flow, because the await didn't seem to terminate.

kgilpin: Since there are various ways this can go wrong, we should probably add telemetry to indicate whether we are trying to do the Python extension for the user - and also stats about how many of our users do and do not already have the Microsoft Python extension. Would you support the idea of disabling the functionality where we go install the Python extension for the user, and then ask them to switch their environments (do they know what this means?) - and just warn/notify if the Python extension is not available?

dustinbyrne:
I've pulled out the logic to perform the install and configure the interpreter based on feedback. If the extension is not installed, we'll just optionally open up the extension in the sidebar without installing it or waiting on it.

kgilpin:
> I've pulled out the logic to perform the install and configure the interpreter based on feedback. If the extension is not installed, we'll just optionally open up the extension in the sidebar without installing it or waiting on it.

OK let's start there and see if telemetry informs us that more is needed.

kgilpin (approved):

PR #709, opened 2023-06-05
feat: Sidebar instructions CTA

kgilpin:
Prompt the user to open the instructions when they don't have any appmap.yml or don't have any AppMap data.

Fixes https://github.com/getappmap/vscode-appland/issues/708

Screen Shot 2023-06-07 at 12 08 30 PM


Daniel-Warner-X (commented):
When I open VS Code to the size I work in, most of the messages are cut off. It feels really noisy..
Screen Shot 2023-06-07 at 12 49 36 PM
It clears up when I expand to full screen, but just barely. There is no room for variation. 
Is it possible to have only the 'Instructions' and 'AppMaps' panes open by default? That gives the message room to breath. 
image

Daniel-Warner-X:
Cool. I see the same thing (with panes being left expanded or closed based on previous actions) but it helps a lot to have them closed to start. The covered content is more tolerable somehow once I have interacted with the sidebar.

PR #714, opened 2023-06-08
Presence of AppMaps is now more accurately presented in the sidebar

### LINKED ISSUE #550: `Record AppMaps` appears as completed in the sidebar despite no AppMaps created
> on a machine that has never run AppMap before

> This is definitely a brand new clone of rails sample app on a machine that hasn't run it before

Image

dustinbyrne:
[Peek 2023-06-08 16-56.webm](https://github.com/getappmap/vscode-appland/assets/8737782/bf029927-d349-4b1c-b0d5-8f43119123bf)

Fixes #550

### DIFF
--- a/src/lib/indexJanitor.ts
+++ b/src/lib/indexJanitor.ts
@@ -1,0 +1,64 @@
+import { Disposable, Uri } from 'vscode';
+import { AppMapWatcher } from '../services/appmapWatcher';
+import { ClassMapWatcher } from '../services/classMapWatcher';
+import { dirname } from 'path';
+import { retry } from '../util';
+import { rm } from 'fs/promises';
+
+enum IndexFile {
+  Metadata = 0,
+  ClassMap,
+  NumFiles,
+}
+
+// IndexJanitor is responsible for deleting index directories once all relevant files within have been deleted.
+// It does so by listening to file change events from AppMapWatcher and ClassMapWatcher and only deleting the index
+// directory once all of the files have been removed. This is necessary because file system watcher events will not
+// be emitted if the index directory is removed before the FileSystemWatcher picks up the file change events.
+//
+// https://github.com/microsoft/vscode/issues/60813
+export default class IndexJanitor implements Disposable {
+  private readonly disposables: Disposable[];
+  private readonly removedIndexFiles = Array.from(
+    { length: IndexFile.NumFiles },
+    () => new Set<string>()
+  );
+
+  constructor(appMapWatcher: AppMapWatcher, classMapWatcher: ClassMapWatcher) {
+    this.disposables = [
+      // AppMapWatcher is actually watching for changes to metadata.json, despite the URI being *.appmap.json.
+      appMapWatcher.onDelete(({ uri }) => this.update(IndexFile.Metadata, uri, 'delete')),
+      appMapWatcher.onCreate(({ uri }) => this.update(IndexFile.Metadata, uri, 'create')),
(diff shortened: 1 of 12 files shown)


src/lib/indexJanitor.ts
```diff
@@ -0,0 +1,65 @@
...
+import { AppMapWatcher } from '../services/appmapWatcher';
+import { ClassMapWatcher } from '../services/classMapWatcher';
+import { dirname } from 'path';
+import { retry } from '../util';
+import { rm } from 'fs/promises';
+
+enum IndexFile {
+  Metadata,
+  ClassMap,
+  NumFiles,
```
dividedmind, line 11: I think this should be renamed to `IndexFileCount` and the enum made to explicitly start at zero to reinforce what the last entry means.

src/lib/indexJanitor.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      currentRemovalState = new Set();
+      this.removedIndexFiles.set(indexPath, currentRemovalState);
+    }
+
+    isRemoved ? currentRemovalState.add(fileType) : currentRemovalState.delete(fileType);
+
+    if (currentRemovalState.size === IndexFile.NumFiles) {
+      this.removedIndexFiles.delete(indexPath);
+      retry(async () => await rm(indexPath, { recursive: true })).catch(console.error);
```
dividedmind, line 56: Both `async` and `await` can be removed here, they're noop in the best case, pure overhead in the worst.

src/lib/indexJanitor.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// IndexJanitor is responsible for deleting index directories once all relevant files within have been deleted.
+// It does so by listening to file change events from AppMapWatcher and ClassMapWatcher and only deleting the index
+// directory once all of the files have been removed. This is necessary because file system watcher events will not
+// be emitted if the index directory is removed before the FileSystemWatcher picks up the file change events.
+//
+// https://github.com/microsoft/vscode/issues/60813
+export default class IndexJanitor implements Disposable {
+  private readonly disposables: Disposable[];
+  private readonly removedIndexFiles: Map<string, Set<IndexFile>> = new Map();
```
dividedmind, line 22: Not that it's terribly important here, but I quite dislike the amount of garbage generated by this data structure. How about two `Set<string>`s for removed classmap and metadata paths? This would avoid repeated creation and disposal of tiny `Set`s. (Alternatively a `Map<string, number>` with bitfields, but that might be a bit of an overkill.)

dividedmind (approved):

src/lib/indexJanitor.ts
dustinbyrne: That's funny, I'd originally implemented it with bitfields. The implementation ended up looking a little too archaic from what I'm used to in TypeScript so I opted to do this instead. 

I switched to use a static array of `Set` instead of the dynamic `Map`.

src/lib/indexJanitor.ts
```diff
@@ -19,15 +19,18 @@ enum IndexFile {
 // https://github.com/microsoft/vscode/issues/60813
 export default class IndexJanitor implements Disposable {
   private readonly disposables: Disposable[];
-  private readonly removedIndexFiles: Map<string, Set<IndexFile>> = new Map();
+  private readonly removedIndexFiles = Array.from(
+    { length: IndexFile.NumFiles },
```
dividedmind, line 23: That's clever! TIL.

PR #717, opened 2023-06-09
Run configs for Java users

ahtrotta:
Fixes getappmap/board#269

This PR helps Java users in VS Code.

* When the extension starts, the latest version of the appmap Java agent JAR will get downloaded to `$HOME/.appmap/java`. This will happen regardless of the language of the project. 

* When a user opens a project, the extension will create a new launch config in `.vscode/launch.json` called `Run with AppMap` that will launch their application with the appmap agent. For this to work, the user must have the [Extension Pack for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack) installed. If they don't, then they will see this message when they attempt to use our launch configuration:

image

* When a user opens a project, the extension will also check to see if they have the [Test Runner for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-test) extension installed (it is part of the Extension Pack for Java). If the user does have it installed, then the extension will add a test config called `Test with AppMap`. If the user already has a test configuration, then when they attempt to run their tests, they will get a drop down menu:

image

* Appmaps will be generated in `tmp/appmap`.

* The extension will only add a launch config once. If the user deletes the launch config, then the extension will not regenerate one. This is also true of the test config.

### DIFF
--- a/src/configuration/extensionState.ts
+++ b/src/configuration/extensionState.ts
@@ -18,6 +18,8 @@ export const Keys = {
     GENERATED_OPENAPI: 'appmap.applandinc.generatedOpenApi',
     HIDE_INSTALL_PROMPT: 'appmap.applandinc.hideInstallPrompt',
     CLOSED_APPMAP: 'appmap.applandinc.closedAppMap',
+    UPDATED_RUN_CONFIG: 'appmap.applandinc.UpdatedLaunchConfig',
+    UPDATED_TEST_CONFIG: 'appmap.applandinc.UpdatedTestConfig',
   },
 };
 
@@ -179,6 +181,22 @@ export default class ExtensionState {
 
   setClosedAppMap(workspaceFolder: vscode.WorkspaceFolder, value: boolean): void {
     this.setWorkspaceFlag(Keys.Workspace.CLOSED_APPMAP, value, workspaceFolder);
+  }
+
+  getUpdatedLaunchConfig(workspaceFolder: vscode.WorkspaceFolder): boolean {
+    return this.getWorkspaceFlag(Keys.Workspace.UPDATED_RUN_CONFIG, workspaceFolder.uri.fsPath);
+  }
+
+  setUpdatedLaunchConfig(workspaceFolder: vscode.WorkspaceFolder, value: boolean): void {
+    this.setWorkspaceFlag(Keys.Workspace.UPDATED_RUN_CONFIG, value, workspaceFolder);
+  }
+
+  getUpdatedTestConfig(workspaceFolder: vscode.WorkspaceFolder): boolean {
+    return this.getWorkspaceFlag(Keys.Workspace.UPDATED_TEST_CONFIG, workspaceFolder.uri.fsPath);
+  }
+
+  setUpdatedTestConfig(workspaceFolder: vscode.WorkspaceFolder, value: boolean): void {
+    this.setWorkspaceFlag(Keys.Workspace.UPDATED_TEST_CONFIG, value, workspaceFolder);
   }
 
   resetState(): void {
--- a/src/services/runConfigService.ts
+++ b/src/services/runConfigService.ts
@@ -1,0 +1,155 @@
+import * as vscode from 'vscode';
+
+import assert from 'assert';
+import path from 'path';
+
+import { WorkspaceService, WorkspaceServiceInstance } from './workspaceService';
+import ProjectStateService, { ProjectStateServiceInstance } from './projectStateService';
+import { WorkspaceServices } from './workspaceServices';
+import ExtensionState from '../configuration/extensionState';
+import { CONFIG_ADDED, DEBUG_EXCEPTION, Telemetry } from '../telemetry';
+import ErrorCode from '../telemetry/definitions/errorCodes';
+
+type JavaTestConfig = {
+  name?: string;
+  vmArgs?: Array<string>;
+};
+
+export class RunConfigServiceInstance implements WorkspaceServiceInstance {
+  private static JAVA_TEST_RUNNER_EXTENSION_ID = 'vscjava.vscode-java-test';
+  protected disposables: vscode.Disposable[] = [];
+
+  public get appmapLaunchConfig(): vscode.DebugConfiguration {
+    return {
+      type: 'java',
+      name: 'Run with AppMap',
+      request: 'launch',
+      mainClass: '',
+      vmArgs: `-javaagent:${this.javaJarPath}`,
+    };
+  }
+
+  public get appmapTestConfig(): JavaTestConfig {
+    return {
+      name: 'Test with AppMap',
+      vmArgs: [`-javaagent:${this.javaJarPath}`],
+    };
+  }
+
+  private get hasPreviouslyUpdatedLaunchConfig(): boolean {
+    return this.extensionState.getUpdatedLaunchConfig(this.folder);
+  }
+
+  private get hasPreviouslyUpdatedTestConfig(): boolean {
+    return this.extensionState.getUpdatedTestConfig(this.folder);
+  }
+
+  private get javaJarPath(): string {
+    return path.join('${userHome}', '.appmap', 'lib', 'java', 'appmap.jar');
+  }
+
+  constructor(
+    public folder: vscode.WorkspaceFolder,
+    public projectStateServiceInstance: ProjectStateServiceInstance,
+    private extensionState: ExtensionState
+  ) {
+    projectStateServiceInstance.onStateChange(this.updateConfigs.bind(this));
+    vscode.extensions.onDidChange(this.updateConfigs.bind(this));
+
+    this.updateConfigs();
+  }
+
+  private async updateConfigs(): Promise<void> {
+    if (!this.isJavaProject()) return;
+    if (!this.hasPreviouslyUpdatedLaunchConfig) await this.updateLaunchConfig();
+    if (!this.hasPreviouslyUpdatedTestConfig) await this.updateTestConfig();
+  }
+
+  public async updateLaunchConfig(): Promise<void> {
+    try {
+      await this.updateConfig('launch', 'configurations', this.appmapLaunchConfig);
+      this.extensionState.setUpdatedLaunchConfig(this.folder, true);
+
+      this.sendConfigUpdateSuccess('launch');
+    } catch (e) {
+      this.sendConfigUpdateError(e);
+    }
+  }
+
+  public async updateTestConfig(): Promise<boolean> {
+    if (this.hasJavaTestExtension()) {
+      try {
+        await this.updateConfig('java.test', 'config', this.appmapTestConfig);
+        this.extensionState.setUpdatedTestConfig(this.folder, true);
+
+        this.sendConfigUpdateSuccess('test');
+        return true;
+      } catch (e) {
+        this.sendConfigUpdateError(e);
+        return false;
+      }
+    }
+    return false;
+  }
+
+  private async updateConfig(
+    parentSection: string,
+    section: string,
+    appmapConfig: vscode.DebugConfiguration | JavaTestConfig
+  ): Promise<void> {
+    const parentConfig = vscode.workspace.getConfiguration(parentSection);
+    let configs = parentConfig.get<Array<vscode.DebugConfiguration | JavaTestConfig>>(section);
+    if (!Array.isArray(configs)) configs = [];
+
+    configs.push(appmapConfig);
+    await parentConfig.update(section, configs);
+  }
+
+  private isJavaProject(): boolean {
+    const { language } = this.projectStateServiceInstance.metadata;
+    return !!(language && language.name && language.name === 'Java');
+  }
+
+  public hasJavaTestExtension(): boolean {
+    const extensions = vscode.extensions.all;
+    return extensions.some(
+      (exension) => exension.id === RunConfigServiceInstance.JAVA_TEST_RUNNER_EXTENSION_ID
+    );
+  }
+
+  private sendConfigUpdateError(e: unknown): void {
+    const err = e as Error;
+    Telemetry.sendEvent(DEBUG_EXCEPTION, {
+      exception: err,
+      errorCode: ErrorCode.ConfigUpdateError,
+    });
+  }
+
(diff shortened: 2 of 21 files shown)


src/services/nodeProcessService.ts
```diff
@@ -163,6 +165,41 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+    const asset = assets.find((a) => a.content_type === 'application/java-archive');
+    if (!asset) throw new Error('Could not retrieve latest Java agent release');
+
+    const javaAgentDir = path.join(os.homedir(), '.appmap', 'java');
+    if (!(await fileExists(javaAgentDir))) mkdirSync(javaAgentDir, { recursive: true });
+
+    const assetPath = path.join(javaAgentDir, asset.name);
+    if (await fileExists(assetPath)) return;
+
+    console.log(`Downloading AppMap Java agent ${asset.name}`);
```
dustinbyrne, line 180: Prefer `log` in this context as it goes to an output window instead of `console.log`

src/services/runConfigService.ts
```diff
@@ -0,0 +1,156 @@
...
+    processService.onReady(this.updateConfigs.bind(this));
+    projectStateServiceInstance.onStateChange(this.updateConfigs.bind(this));
+
+    if (processService.ready) this.updateConfigs();
+  }
+
+  private async updateConfigs(): Promise<void> {
+    if (!this.shouldUpdate()) return;
+
+    if (!this.hasPreviouslyUpdatedLaunchConfig) {
```
dustinbyrne, line 62: I had an issue where I needed to recreate the `launch.json` but this prevented me from doing so easily. I had to `AppMap: Reset Usage State` to get it working. Maybe an issue for the future, but it'd be useful to have some easy/discoverable way of resetting this.

kgilpin:
How do you feel about $HOME/.appmap/lib ? Would be a bit more standard, especially if we add other language libraries and bin scripts.

src/services/nodeProcessService.ts
```diff
@@ -163,6 +165,41 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+    } catch (e) {
+      // if the symlink that we're trying to remove does not exist, don't do anything
+    }
+
+    let symlinkCreated = true;
+    try {
+      symlinkSync(assetPath, symlinkPath, 'file');
+      this.extensionState.setCurentJavaJarPath(symlinkPath);
+    } catch (e) {
+      this.extensionState.setCurentJavaJarPath(assetPath);
```
dustinbyrne, line 195: `Current`

ahtrotta:
> How do you feel about $HOME/.appmap/lib ? Would be a bit more standard, especially if we add other language libraries and bin scripts.

I'm fine with `$HOME/.appmap/lib` but I think I'd want it to be `$HOME/.appmap/lib/java` so the symlink would be `$HOME/.appmap/lib/java/latest` instead of `$HOME/.appmap/lib/latest` because the latter is ambiguous.

src/services/nodeProcessService.ts
ahtrotta: Okee dokee, I changed it :+1:

src/telemetry/definitions/properties.ts
```diff
@@ -359,3 +359,24 @@ export const LINK_TYPE = new TelemetryDataProvider({
     return linkType;
   },
 });
+
+export const SYMLINK_CREATED = new TelemetryDataProvider({
+  id: 'appmap.symlink_created',
+  async value({ symlinkCreated }: { symlinkCreated: boolean }) {
+    return symlinkCreated;
```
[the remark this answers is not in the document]
ahtrotta: Good catch!

src/services/nodeProcessService.ts
ahtrotta: :upside_down_face:

kgilpin:
> > How do you feel about $HOME/.appmap/lib ? Would be a bit more standard, especially if we add other language libraries and bin scripts.
> 
> I'm fine with `$HOME/.appmap/lib` but I think I'd want it to be `$HOME/.appmap/lib/java` so the symlink would be `$HOME/.appmap/lib/java/latest` instead of `$HOME/.appmap/lib/latest` because the latter is ambiguous.

Can you use appmap.jar as the latest version? Otherwise the VM configuration is going to be weird looking. A user will be expect the javaagent argument to be a JAR file name.

src/services/runConfigService.ts
ahtrotta: > > > How do you feel about $HOME/.appmap/lib ? Would be a bit more standard, especially if we add other language libraries and bin scripts.
> > 
> > 
> > I'm fine with `$HOME/.appmap/lib` but I think I'd want it to be `$HOME/.appmap/lib/java` so the symlink would be `$HOME/.appmap/lib/java/latest` instead of `$HOME/.appmap/lib/latest` because the latter is ambiguous.
> 
> Can you use appmap.jar as the latest version? Otherwise the VM configuration is going to be weird looking. A user will be expect the javaagent argument to be a JAR file name.

Yeah, good point

ahtrotta:
> > > How do you feel about $HOME/.appmap/lib ? Would be a bit more standard, especially if we add other language libraries and bin scripts.
> > 
> > 
> > I'm fine with `$HOME/.appmap/lib` but I think I'd want it to be `$HOME/.appmap/lib/java` so the symlink would be `$HOME/.appmap/lib/java/latest` instead of `$HOME/.appmap/lib/latest` because the latter is ambiguous.
> 
> Can you use appmap.jar as the latest version? Otherwise the VM configuration is going to be weird looking. A user will be expect the javaagent argument to be a JAR file name.

Yeah, good point

src/services/nodeProcessService.ts
```diff
@@ -163,6 +165,41 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
...
+    } catch (e) {
+      // if the symlink that we're trying to remove does not exist, don't do anything
+    }
+
+    let symlinkCreated = true;
+    try {
+      symlinkSync(assetPath, symlinkPath, 'file');
+      this.extensionState.setCurrentJavaJarPath(symlinkPath);
+    } catch (e) {
+      this.extensionState.setCurrentJavaJarPath(assetPath);
```
kgilpin, line 195: If you can't create the symlink, just copy the versioned JAR to `appmap.jar`.

src/configuration/extensionState.ts
```diff
@@ -181,11 +184,35 @@ export default class ExtensionState {
...
+  getUpdatedTestConfig(workspaceFolder: vscode.WorkspaceFolder): boolean {
+    return this.getWorkspaceFlag(Keys.Workspace.UPDATED_TEST_CONFIG, workspaceFolder.uri.fsPath);
+  }
+
+  setUpdatedTestConfig(workspaceFolder: vscode.WorkspaceFolder, value: boolean): void {
+    this.setWorkspaceFlag(Keys.Workspace.UPDATED_TEST_CONFIG, value, workspaceFolder);
+  }
+
+  getCurrentJavaJarPath(): string | undefined {
+    return this.context.globalState.get(Keys.Global.CURRENT_JAVA_JAR);
```
kgilpin, line 204: This should always be `$HOME/.appmap/lib/java/appmap.jar`. See my comments later on.

src/telemetry/definitions/events.ts
```diff
@@ -214,3 +214,13 @@ export const INSTALL_BUTTON_ABORT = new Event({
   name: 'install_button:abort',
   properties: [Properties.REASON, Properties.OPTIONAL_RESULT],
 });
+
+export const DOWNLOADED_JAVA_JAR = new Event({
+  name: 'downloaded_java_jar',
+  properties: [Properties.SYMLINK_CREATED],
```
kgilpin, line 220: Include the new version installed as a property.

src/telemetry/definitions/events.ts
```diff
@@ -214,3 +214,13 @@ export const INSTALL_BUTTON_ABORT = new Event({
...
+});
+
+export const CONFIG_ADDED = new Event({
+  name: 'config_added',
+  properties: [Properties.CONFIG_TYPE, Properties.JAR_PATH],
```
kgilpin, line 225: JAR_PATH can be omitted because it's always the same.

src/telemetry/definitions/properties.ts
```diff
@@ -359,3 +359,24 @@ export const LINK_TYPE = new TelemetryDataProvider({
...
+});
+
+export const CONFIG_TYPE = new TelemetryDataProvider({
+  id: 'appmap.config_type',
+  async value({ configType }: { configType: string }) {
+    return configType;
+  },
+});
+
+export const JAR_PATH = new TelemetryDataProvider({
```
kgilpin, line 377: Can be dropped IMO.

test/integration/runConfigs/runConfigsJava.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    await waitFor('Waiting for dependency installation', () => processService.ready);
+  });
+
+  after(async () => {
+    sinon.restore();
+    await initializeWorkspace();
+  });
+
+  it('downloads the latest jar', () => {
```
kgilpin, line 55: You may want to separate the JAR file download tests from the run config tests, just for organizational purposes. At least, into different contexts.

test/integration/runConfigs/runConfigsJava.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    assert(existsSync(ExpectedJarPath(ProjectJava)));
```
kgilpin, line 56: It would be great to see tests here that cover:

a) Symlink command fails and so the extension falls back to copying the JAR
b) Download lock file exists so the JAR update is a no-op.

test/integration/runConfigs/runConfigsRuby.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    runConfigServiceInstance = runConfigServiceInstances[0];
+
+    await waitFor('Waiting for dependency installation', () => processService.ready);
+  });
+
+  after(async () => {
+    sinon.restore();
+    await initializeWorkspace();
+  });
+
```
kgilpin, line 52: You probably want a context function here like `in a Ruby project`

src/services/nodeProcessService.ts
ahtrotta: Yup, good idea, I implemented this

src/services/nodeProcessService.ts
```diff
@@ -163,6 +165,41 @@ export class NodeProcessService implements WorkspaceService<NodeProcessServiceIn
     return await downloadFile(url, cliBin);
   }
 
+  async getLatestJavaJar(log: (line: string) => void): Promise<void> {
```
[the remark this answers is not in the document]
ahtrotta: Yup, I implemented this, too

src/configuration/extensionState.ts
ahtrotta: :+1:

src/telemetry/definitions/events.ts
ahtrotta: :+1:

src/telemetry/definitions/events.ts
ahtrotta: :+1:

src/telemetry/definitions/properties.ts
ahtrotta: Done

test/integration/runConfigs/runConfigsJava.test.ts
ahtrotta: Done

test/integration/runConfigs/runConfigsJava.test.ts
ahtrotta: a) I spent probably way too long trying to stub `symlinkSync` and make it throw an error and I never got it to work, so there isn't a test for this at the moment
b) I added a test for this

test/integration/runConfigs/runConfigsRuby.test.ts
ahtrotta: I changed the `describe` string

src/services/runConfigService.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    configs.push(appmapConfig);
+    await parentConfig.update(section, configs);
+  }
+
+  private shouldUpdate(): boolean {
+    const hasJavaJar = existsSync(this.javaJarPath);
+    const isReady = this.processService.ready;
+    const { language } = this.projectStateServiceInstance.metadata;
+    const isJavaProject = language && language.name && language.name === 'Java';
```
kgilpin, line 113: I feel like `isJavaProject` is the only condition that's really needed. It seems that we can just keep it simple and update the run configs for Java projects. Whether the process service is ready and whether the JAR is installed seem like separate concerns, if we have too many checks then we risk the launch config not getting updated and we aren't sure why - race condition? Weird project metadata problem? If we always update the Java run configs, and say the JAR file isn't installed, then at least the user can see the intent and ask us to help sort it out.

src/services/runConfigService.ts
```diff
@@ -0,0 +1,166 @@
...
+      this.extensionState.setUpdatedLaunchConfig(this.folder, true);
+
+      this.sendConfigUpdateSuccess('launch');
+    } catch (e) {
+      this.sendConfigUpdateError(e);
+    }
+  }
+
+  public async updateTestConfig(): Promise<void> {
+    if (this.hasJavaTestExtension()) {
```
kgilpin, line 84: Do we care about `hasJavaTestExtension` or can we just look for the `java.test.config` and update it if exists?

dustinbyrne: I think for our purposes the two are synonymous as `java.test.config` is a property introduced by the Test Runner for Java extension.

ahtrotta: Yeah, if we try to modify `java.test.config` and the `Test Runner for Java` extension is not installed, then VS Code will throw an error, so we need to check if that extension is installed first.

kgilpin (approved):

PR #720, opened 2023-06-15
fix: Project language now reflects the preferred language

### LINKED ISSUE #718: Telemetry for reported language should match what the Project Picker determines
This should not be returning/stringifying an array:
https://github.com/getappmap/vscode-appland/blob/master/src/telemetry/definitions/properties.ts#L149

Instead, return the single language we'd be displaying in the project picker.

dustinbyrne:
This PR fixes an issue where project language was being reported as an array of all languages present in the workspace folder. This presented an issue, because there was now no indication of what language the project would be defined as by the project picker.

Resolves #718

### DIFF
--- a/src/telemetry/definitions/properties.ts
+++ b/src/telemetry/definitions/properties.ts
@@ -16,6 +16,9 @@ import {
   AppmapConfigManager,
   AppmapConfigManagerInstance,
 } from '../../services/appmapConfigManager';
+import ProjectStateService, {
+  ProjectStateServiceInstance,
+} from '../../services/projectStateService';
 
 export const DEBUG_EXCEPTION = new TelemetryDataProvider({
   id: 'appmap.debug.exception',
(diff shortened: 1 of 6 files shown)


src/telemetry/definitions/properties.ts
```diff
@@ -148,8 +151,12 @@ export const SCANNER_CONFIG_PRESENT = new TelemetryDataProvider({
 
 export const PROJECT_LANGUAGE = new TelemetryDataProvider({
   id: 'appmap.project.language',
-  cache: true,
-  async value(data: { rootDirectory?: string; metadata?: Record<string, unknown> }) {
+  cache: false,
+  async value(data: {
```
dividedmind, line 155: You can remove the useless `async` while you're here.

dividedmind (approved):
Looks ok.

The other commit (with the `initialized` parameter to `onCreate`) seems unrelated and unused, did you mean to include it?

PR #721, opened 2023-06-16
feat: Reuse existing terminals when rerunning installer if possible

### LINKED ISSUE #703: Reuse the same terminal for each run of the Install command
Reuse the same terminal for each invocation of the Install command, so that the user has a chance to modify the environment and retry.

dividedmind:
Fixes #703

### DIFF
--- a/src/lib/dictEqual.ts
+++ b/src/lib/dictEqual.ts
@@ -1,0 +1,14 @@
+export type Dict = { [key: string]: string | null | undefined | number };
+
+/** Compares two dictionary-like objects for equality. */
+export default function dictEqual(xs: Dict | undefined, ys: Dict | undefined): boolean {
+  if (xs === ys) return true;
+  if (!(xs && ys)) return false;
+
+  const xse = Object.entries(xs);
+  const yse = Object.entries(ys);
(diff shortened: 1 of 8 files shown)


src/lib/dictEqual.ts
```diff
@@ -0,0 +1,14 @@
...
+export default function dictEqual(xs: Dict | undefined, ys: Dict | undefined): boolean {
+  if (xs === ys) return true;
+  if (!(xs && ys)) return false;
+
+  const xse = Object.entries(xs);
+  const yse = Object.entries(ys);
+  if (xse.length !== yse.length) return false;
+
+  for (const [k, v] of xse) if (ys[k] !== v) return false;
+  return true;
```
ahtrotta, line 13: Can you make the names in this function more readable?

`xs`: `dictA`
`ys`: `dictB`
`xse`: `dictAEntries`
`yse`: `dictBEntries`
`k`: `key`
`v`: `value`

You don't have to use those exactly, but something along those lines.

ahtrotta: Okee dokee

dustinbyrne (approved):

PR #728, opened 2023-06-21
Users can update an existing filter

### LINKED ISSUE #727: Users can update an existing saved filter
Currently, when a user saves a filter with the same name as a previously saved filter there end up being two filters with the exact same name. Then, when the user tries to delete one of them, they both get deleted.

ahtrotta:
Fixes #727 

Fixes a bug where users could not update an existing saved filter.

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -216,10 +216,19 @@ export default class AppMapEditorProvider
   }
 
   async saveFilter(filter: SavedFilter): Promise<void> {
-    let savedFilters = this.getSavedFilters();
-    if (!savedFilters) savedFilters = [];
+    let savedFilters = this.getSavedFilters() || [];
 
-    savedFilters.push(filter);
+    let filterOverwritten = false;
+    savedFilters = savedFilters.map((savedFilter) => {
+      if (savedFilter.filterName === filter.filterName) {
+        filterOverwritten = true;
+        return filter;
+      }
+      return savedFilter;
+    });
+
+    if (!filterOverwritten) savedFilters.push(filter);
+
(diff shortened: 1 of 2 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -219,7 +219,17 @@ export default class AppMapEditorProvider
     let savedFilters = this.getSavedFilters();
```
kgilpin, line 219: `this.getSavedFilters() || []`

src/editor/appmapEditorProvider.ts
```diff
@@ -219,7 +219,17 @@ export default class AppMapEditorProvider
...
     if (!savedFilters) savedFilters = [];
```
kgilpin, line 220: `let savedFilters = this.getSavedFilters() || []`

test/integration/savedFilters.test.ts
```diff
@@ -53,6 +53,22 @@ describe('Saved filters', () => {
...
+    const updatedFilter = {
+      filterName: 'test',
+      state: 'eyJmaWx0ZXJzIjp7ImxpbWl0Um9vdEV2ZW50cyI6ZmFsc2UsImhpZGVVbmxhYmVsZWQiOnRydWV9fQ',
+      default: false,
+    };
+
+    await editorProvider.saveFilter(updatedFilter);
+    const savedFilters = editorProvider.getSavedFilters();
+
+    assert.deepEqual(savedFilters, [defaultFilter, updatedFilter]);
```
kgilpin, line 68: It would be a bit more clear if you also make this same assertion before adding `updatedFilter`, then it would be super obvious that you're asserting that the filter list hasn't changed in length.

kgilpin (approved):

PR #733, opened 2023-06-23
Java test config respects appmap_dir

### LINKED ISSUE #730: JUnit AppMaps not appearing in the AppMap list
AppMaps from the junit test are getting created in `tmp/appmap/junit` but they are not showing up in the IDE extension.

via @petecheslock

Image

Image

ahtrotta:
Fixes #730

- [x] When a test config is created, a vmArg is added that specifies the output directory based on the current value of `appmap_dir` in `appmap.yml`
- [x] Checks whether there is a `Test with AppMap` test config without the required `-Dappmap.output.directory` vmArg and adds the vmArg if it is not present.
- [x] If there is no `appmap_dir` in `appmap.yml` then the default is `target/appmap`.
- [x] Also, removes the intermittently failing test `creates a new test config` and adds an equivalent unit test in
`runConfigs/testConfigsJava.test.ts`.

### DIFF
--- a/src/commands/getAppmapDir.ts
+++ b/src/commands/getAppmapDir.ts
@@ -1,0 +1,38 @@
+import * as vscode from 'vscode';
+
+import assert from 'assert';
+import { WorkspaceServices } from '../services/workspaceServices';
+import { AppmapConfigManager, AppmapConfigManagerInstance } from '../services/appmapConfigManager';
+
+export default async function getAppmapDir(
+  context: vscode.ExtensionContext,
+  workspaceServices: WorkspaceServices
+): Promise<void> {
+  const command = vscode.commands.registerCommand(
+    'appmap.getAppmapDir',
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async (arg: any): Promise<string> => {
+      const { cwd } = arg;
+      // AFAIK this will always have cwd, but we'll check just in case
+      if (!cwd) return 'tmp/appmap';
+
+      const uri = vscode.Uri.file(cwd);
+      const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri);
+      assert(workspaceFolder);
+
+      const configInstance = workspaceServices.getServiceInstanceFromClass(
+        AppmapConfigManager,
+        workspaceFolder
+      ) as AppmapConfigManagerInstance | undefined;
+      assert(configInstance);
+
+      const config = await configInstance.getAppmapConfig();
+      // TODO: determine project type and pick default based on that
+      if (!config) return 'tmp/appmap';
+
+      return config.appmapDir;
+    }
+  );
+
+  context.subscriptions.push(command);
+}
--- a/src/services/runConfigService.ts
+++ b/src/services/runConfigService.ts
@@ -17,6 +17,8 @@ type JavaTestConfig = {
 
 export class RunConfigServiceInstance implements WorkspaceServiceInstance {
   private static JAVA_TEST_RUNNER_EXTENSION_ID = 'vscjava.vscode-java-test';
+  private outputDirVmarg = '-Dappmap.output.directory=${command:appmap.getAppmapDir}';
+  private testConfigName = 'Test with AppMap';
   protected disposables: vscode.Disposable[] = [];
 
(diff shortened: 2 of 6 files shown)


src/commands/getAppmapDir.ts
```diff
@@ -0,0 +1,38 @@
...
+  context: vscode.ExtensionContext,
+  workspaceServices: WorkspaceServices
+): Promise<void> {
+  const command = vscode.commands.registerCommand(
+    'appmap.getAppmapDir',
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async (arg: any): Promise<string> => {
+      const { cwd } = arg;
+      // AFAIK this will always have cwd, but we'll check just in case
+      if (!cwd) return 'target/appmap';
```
dustinbyrne, line 17: `tmp/appmap` should be the default

src/commands/getAppmapDir.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      const configInstance = workspaceServices.getServiceInstanceFromClass(
+        AppmapConfigManager,
+        workspaceFolder
+      ) as AppmapConfigManagerInstance | undefined;
+      assert(configInstance);
+
+      const config = await configInstance.getAppmapConfig();
+      // TODO: determine project type and pick default based on that
+      if (!config) return 'target/appmap';
```
dustinbyrne, line 31: Here too

src/services/runConfigService.ts
```diff
@@ -92,6 +98,29 @@ export class RunConfigServiceInstance implements WorkspaceServiceInstance {
...
+      try {
+        const parentConfig = vscode.workspace.getConfiguration('java.test');
+        const configs = parentConfig.get<Array<JavaTestConfig>>('config');
+        if (!configs || !Array.isArray(configs)) return;
+
+        configs.forEach((config) => {
+          if (
+            config.name === this.testConfigName &&
+            config.vmArgs &&
+            !config.vmArgs.includes(this.outputDirVmarg)
```
dustinbyrne, line 112: It might be better to look for the system property instead of the exact string we'd like to set it to in case the user changes it. E.g.:
```ts
!config.vmArgs.some((arg) => arg.includes('appmap.output.directory'))
```

src/commands/getAppmapDir.ts
ahtrotta: Thanks, fixed it :+1:

src/commands/getAppmapDir.ts
ahtrotta: Thanks, fixed it :+1:

src/services/runConfigService.ts
ahtrotta: Good idea :+1:

dustinbyrne (approved):

src/commands/getAppmapDir.ts
```diff
@@ -0,0 +1,38 @@
...
+import assert from 'assert';
+import { WorkspaceServices } from '../services/workspaceServices';
+import { AppmapConfigManager, AppmapConfigManagerInstance } from '../services/appmapConfigManager';
+
+export default async function getAppmapDir(
+  context: vscode.ExtensionContext,
+  workspaceServices: WorkspaceServices
+): Promise<void> {
+  const command = vscode.commands.registerCommand(
+    'appmap.getAppmapDir',
```
kgilpin, line 12: I thought this had to be registered in `package.json`?

src/services/runConfigService.ts
```diff
@@ -17,6 +17,8 @@ type JavaTestConfig = {
 
 export class RunConfigServiceInstance implements WorkspaceServiceInstance {
   private static JAVA_TEST_RUNNER_EXTENSION_ID = 'vscjava.vscode-java-test';
+  private outputDirVmarg = '-Dappmap.output.directory=${command:appmap.getAppmapDir}';
```
kgilpin, line 20: I'm wishing this was `getAppMapDir`

src/commands/getAppmapDir.ts
ahtrotta: Yeah, I believe that's just necessary for user visibility

kgilpin (commented):
Nice to see this fixed 👍

PR #736, opened 2023-06-23
feat: Update Java install instructions

### LINKED ISSUE #732: Java installation instructions in VS Code doesn't reflect the new "Run with AppMap" flow
The current prompt is to use the CLI to install the Maven or Gradle plugin. Using the downloaded jar file, this is no longer necessary. The user just needs an appmap.yml file.
Image

dustinbyrne:
This pull request introduces changes to lockfile handling and reactive status updates of Java assets.

**Refactoring Lockfile Logic**
The lockfile logic used by the `NodeDependencyService` has been moved into a new class, `LockfileSynchronizer`. This allows us to more easily test and use the logic elsewhere, such as when downloading Java assets.

**Reactive Run Configuration Status**
The PR adds a reactive run configuration status to the project metadata. This allows for more interactive status updates and real-time feedback, which improves usability and user experience.

**Reactive Status Events for Java Asset Downloads**
This PR also introduces reactive status events for Java asset downloads. This includes a name change from `AssetManager` to `JavaAssets`, and `JavaAssetDownloader`. Additionally, it removes the previously used locking mechanism in favor of the same mechanism that's been described above. This change will make the download process more consistent across different types of dependencies and more resilient.

**Integrating Java Asset Status in Instructions**
Lastly, this PR integrates Java asset status into the instructions, providing users with real-time status updates. This feature increases transparency and improves the overall user experience.

See the [front-end pull request](https://github.com/getappmap/appmap-js/pull/1243) for screenshots.

Resolves #732


web/src/installGuideView.js
```diff
@@ -97,11 +100,20 @@ export default function mountInstallGuide() {
...
+    app.$on('view-output', () => {
+      vscode.postMessage({ command: 'view-output' });
+    });
+
     messages.on('page', ({ page }) => {
       app.$refs.ui.jumpTo(page);
     });
 
     messages.on('projects', ({ projects }) => {
+      console.log(projects);
```
ahtrotta, line 116: Is this intentional?

web/src/installGuideView.js
```diff
@@ -51,6 +53,7 @@ export default function mountInstallGuide() {
         });
       },
       mounted() {
+        console.log(this);
```
ahtrotta, line 56: Is this intentional?

ahtrotta (changes requested):
Couple of small things (these are really in `appmap-js`):

image

1. The location of the java agent jar file is `~/.appmap/lib/java/appmap.jar` not `~/.appmap/lib/appmap.jar`.
2. The java agent argument is `-javaagent:` not `-Djavaagent=`

Also, it might be nice to add a button here to attempt to download the java agent jar, similar to how you have a button to re-add the run configs:

image

There's a command for it: `src/commands/downloadLatestJavaJar.ts`.

In the case where there is a test config but not a launch config or a launch config but not a test config, when a user presses the button to re-add them, they get a duplicate of the one that already existed.

For example, if I delete the test config and then use the button in the instructions, I get this launch config:

image

Could we detect if one or both is missing and add the correct one(s)?

I think the most important issue here is that we're adding duplicate test/launch configs if one already exists and the user attempt to re-add them using the button. The text issues can be fixed as a follow on and the button to re-add the java agent jar is optional IMO.

web/src/installGuideView.js
dustinbyrne: Addressed in c554715efa3963972fb0e59072225bc09c392ad8

dustinbyrne: Addressed in https://github.com/getappmap/vscode-appland/pull/736/commits/bba5f10864f6efc70ab810ff8f5b2784b36afd81

web/src/installGuideView.js
dustinbyrne: Addressed in https://github.com/getappmap/vscode-appland/pull/736/commits/bba5f10864f6efc70ab810ff8f5b2784b36afd81

dustinbyrne:
Thanks! I've addressed the concerns with the duplicate launch/test config entries in https://github.com/getappmap/vscode-appland/pull/736/commits/556b458a163f195912c1785ab21aef5f556389ee. I've updated the frontend in https://github.com/getappmap/vscode-appland/pull/736/commits/caa02b818657721a839a7c82002b98ce990ac91d to address the frontend concerns.

Regarding the button to run `AppMap: Update Java Jar`, I think it's valid feedback and I can see its usefulness. For now, I'll open a separate issue to avoid holding up the delivery of this PR any longer.

PR #737, opened 2023-06-26
refactor: Asset manager

kgilpin:
Refactor Asset Manager and make the tests simpler and more reliable.


src/lib/assetManager.ts
```diff
@@ -0,0 +1,115 @@
+import {
```
kgilpin, line 1: This code is primarily moved here from `src/services/assetManager.ts`. Note that it doesn't import or use anything from `vscode`.

src/lib/assetManager.ts
```diff
@@ -0,0 +1,115 @@
...
+  symlinkSync,
+  unlinkSync,
+  writeFileSync,
+} from 'fs';
+import os from 'os';
+import path from 'path';
+
+import GithubRelease, { GithubReleaseAsset } from '../lib/githubRelease';
+
+export interface ProgressReporter {
```
kgilpin, line 16: Install progress is reported in a structured way through this interface. That way it's not necessary to make assertions about log messages - which is fragile. It's very easy to see and assert exactly what the AssetManager is doing by monitoring these function calls.

src/services/assetManager.ts
```diff
@@ -1,104 +1,58 @@
...
-import { fileExists } from '../util';
 import ErrorCode from '../telemetry/definitions/errorCodes';
+import AssetManager, { ProgressReporter } from '../lib/assetManager';
 
-export default class AssetManager {
-  private static FIVE_MINUTES_IN_MILLISECONDS = 5 * 60 * 1000;
-  private static appmapJavaRelease = new GithubRelease('getappmap', 'appmap-java');
-  private static outputChannel = vscode.window.createOutputChannel('AppMap: Assets');
-  public static logLines: string[] = [];
+class ProgressReporterImpl implements ProgressReporter {
```
kgilpin, line 8: The log statements that were in the existing implementation are still here. But they are not important from an assertion standpoint.

src/services/assetManager.ts
```diff
@@ -1,104 +1,58 @@
...
 
-  public static async register(): Promise<void> {
-    await this.getLatestJavaJar();
+  protected log(line: string) {
+    this.logLines.push(line);
+    if (this.outputChannel) this.outputChannel.appendLine(line);
   }
 }
+
+export default async function installLatestJavaJar(raiseOnError: boolean) {
```
kgilpin, line 55: When run in the background, errors are swallowed and reported via telemetry. When run as a user command, the exception is handled by the caller.

PR #738, opened 2023-06-26
feat: Organize runtime analysis

kgilpin:
* Organize findings by date.
* Include failed tests in the findings view.
* Include inline commands to view AppMap as JSON, and delete an AppMap.

Screen Shot 2023-06-26 at 4 33 06 PM


src/extension.ts
```diff
@@ -33,7 +33,6 @@ import appmapLinkProvider from './terminalLink/appmapLinkProvider';
 import registerTrees from './tree';
 import { ClassMapTreeDataProvider } from './tree/classMapTreeDataProvider';
 import ContextMenu from './tree/contextMenu';
-import { FindingsTreeDataProvider } from './tree/findingsTreeDataProvider';
```
kgilpin, line 36: Moved to src/tree/index.ts

src/services/findingsIndex.ts
```diff
@@ -15,22 +15,22 @@ export default class FindingsIndex extends EventEmitter implements vscode.Dispos
   private findingsBySourceUri = new Map<string, ResolvedFinding[]>();
```
kgilpin, line 15: This property is a `Map` but it's used like an `Object` throughout this class.

test/fixtures/workspaces/project-several-findings/.gitignore
```diff
@@ -1 +1,3 @@
 node_modules
```
kgilpin, line 1: The rest of the file changes are all tests.

test/fixtures/workspaces/project-several-findings/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,96 @@
+class UsersController < ApplicationController
```
kgilpin, line 1: These fixture source files need to be exist in order for the finding problem location to resolve.

test/fixtures/workspaces/project-several-findings/appmap.yml
```diff
@@ -1 +1,5 @@
-# dummy
\ No newline at end of file
+name: project-several-findings
```
kgilpin, line 1: Instead of having static fixture findings, this project is now actually scanned.

test/integration/analysisTree/enumerateTree.ts
```diff
@@ -0,0 +1,27 @@
+import * as vscode from 'vscode';
```
kgilpin, line 1: The runtime analysis test will basically assert that the state of the entire tree matches an expectation, which is loaded from a JSON file. All that's tested is the labels, but that's enough to verify that it's working as expected.

src/lib/deleteAppMap.ts
```diff
@@ -13,11 +14,14 @@ export default async function deleteAppMap(uri: vscode.Uri): Promise<void> {
...
   await retry(async () => rm(uri.fsPath, { force: true }));
-  await Promise.all(
-    filesToDelete.map((file) => retry(async () => await rm(file, { force: true })))
-  );
+  await retry(async () => {
+    for (const file of await promisify(glob)(`${indexDir}/*`)) {
+      await rm(file, { force: true });
+    }
+  });
+  await retry(async () => rm(indexDir, { force: true, recursive: true }));
```
dustinbyrne, line 26: We shouldn't remove the index directory here because it's likely that we'll miss `FileWatcher` events for files contained within. This is a limitation of VSCode's API (https://github.com/microsoft/vscode/issues/60813). There's now a separate service responsible for removing index directories once the contents have been ACK'd as deleted ([`IndexJanitor`](https://github.com/getappmap/vscode-appland/blob/49b8eb16964422110eafc8ce92d73d687a7801f3/src/lib/indexJanitor.ts#L20)).

src/tree/findingsTreeDataProvider.ts
```diff
@@ -38,110 +172,151 @@ export class FindingsTreeDataProvider
...
 
-        if (impactDomain && !impactDomains.includes(impactDomain)) {
-          impactDomains.push(impactDomain);
-        }
+    if (this.appmaps.appMaps().length === 0) return [];
 
-        return impactDomains;
-      }, [] as string[])
-      .sort((labelA, labelB) => IMPACT_DOMAINS.indexOf(labelA) - IMPACT_DOMAINS.indexOf(labelB));
+    const projects = vscode.workspace.workspaceFolders || [];
```
dustinbyrne, line 186: Nit - you shouldn't need to coerce this into an array because `undefined` has already been ruled out above

src/tree/findingsTreeDataProvider.ts
```diff
@@ -38,110 +172,151 @@ export class FindingsTreeDataProvider
...
+    const overviewTreeItem = new vscode.TreeItem('Findings table');
     overviewTreeItem.command = {
       command: 'appmap.openFindingsOverview',
       title: `Open in AppMap`,
     };
     overviewTreeItem.iconPath = new vscode.ThemeIcon('preview');
 
-    const topLevelTreeItems = [overviewTreeItem];
+    const projectsWithFindings = this.findingsIndex
+      ?.findings()
```
dustinbyrne, line 196: Similarly the optional chain isn't necessary because `this.findingsIndex` is guaranteed to be defined. Normally the optional chain would cause the resulting type to optionally be `undefined` but in this case it looks like TypeScript is smart enough to recognize it'll always be defined.

dustinbyrne (changes requested):
Looks great! The only necessary change from my perspective is to retain the index directory (https://github.com/getappmap/vscode-appland/pull/738#discussion_r1243994316). Other than that, it seems ready to go out.

src/lib/deleteAppMap.ts
kgilpin: OK. I was getting a test error related to the presence, or absence (I forget which) of this directory. I'll remove this line and if I see that error again I will send it over to you :-) I'm 99% sure it was unrelated to the changes in this PR.

dustinbyrne (approved):

PR #741, opened 2023-06-28
Appmap file watchers respect `appmap_dir`

### LINKED ISSUE #740: The AppMap watcher looks for appmaps according to the `appmap_dir` in the `appmap.yml`
Currently, the appmap watcher in VS Code is hard-coded to look in `tmp/appmap`, `build/appmap`, and `target/appmap`, but if the user specifies another location as their `appmap_dir`, then the AppMaps will not show up in the sidebar.

ahtrotta:
Fixes #740


kgilpin:
No offense, but this doesn’t actually fix any real work problem.

PR #743, opened 2023-06-29
Open appmap sign-in on installation

### LINKED ISSUE #742: Open sign-in page when extension is installed

ahtrotta:
Fixes #742 

This change makes it so that the appmap sidebar is opened to the sign-in page when the extension is installed. 

If the user is already signed-in, then the sign-in page won't be visible and the command will do nothing.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -292,6 +292,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
 
     await AnalysisManager.register(context, projectStates, extensionState, workspaceServices);
 
+    vscode.commands.executeCommand('appmap.views.signIn.focus');
+
     return {
       editorProvider,
       localAppMaps: appmapCollectionFile,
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -292,6 +292,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
 
     await AnalysisManager.register(context, projectStates, extensionState, workspaceServices);
 
+    vscode.commands.executeCommand('appmap.views.signIn.focus');
+
```
dustinbyrne, line 296: This will open up for anyone who isn't signed in. Can we limit it to new installs?

dustinbyrne (changes requested):

src/extension.ts
ahtrotta: Sure! I'll change it.

dustinbyrne (approved):

PR #745, opened 2023-06-29
AppMaps are organized by workspace in tree view

### LINKED ISSUE #744: AppMaps in the sidebar are organized by workspace
Currently, if multiple workspaces are open then all the appmaps in the sidebar are mixed together. They should be separated by workspace, similar to how @kgilpin just organized the findings.

ahtrotta:
Fixes #744 

Using a workspace with both `spring-petclinic` and `sample_app_6th_ed` open:

Before:

image

After:

image

### DIFF
--- a/test/integration/appmaps/appmapsTree.test.ts
+++ b/test/integration/appmaps/appmapsTree.test.ts
@@ -8,29 +8,38 @@ describe('AppMaps', () => {
   beforeEach(waitForExtension);
   afterEach(initializeWorkspace);
 
-  it('is a two-level tree', async () => {
+  it('is a three-level tree', async () => {
     const trees = (await waitForExtension()).trees;
 
     const appmapsTree = trees.appmaps;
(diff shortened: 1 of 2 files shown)


test/integration/appmaps/appmapsTree.test.ts
```diff
@@ -8,29 +8,38 @@ describe('AppMaps', () => {
   beforeEach(waitForExtension);
   afterEach(initializeWorkspace);
 
-  it('is a two-level tree', async () => {
+  it.only('is a three-level tree', async () => {
```
kgilpin, line 11: What does `only` do?

ahtrotta: `only` makes it so that only that test runs, which is a remnant from my local testing. I'll remove it!

dustinbyrne (approved):

PR #762, opened 2023-07-13
feat: Update the Java installation condition

dustinbyrne:
When the agent and run configurations exist, the Java installation will be flagged as complete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -499,7 +499,7 @@
   "dependencies": {
     "@appland/appmap": "^3.80.2",
     "@appland/client": "^1.7.0",
-    "@appland/components": "^2.55.0",
+    "@appland/components": "^2.56.0",
     "@appland/diagrams": "^1.7.0",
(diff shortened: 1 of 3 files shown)


ahtrotta:
Cool, I just tested it manually and removed my launch config and test config and it reacted to both changes perfectly. I noticed that when I deleted my `~/.appmap` folder the instructions page didn't update, though:

image

We would need a file watcher for this I would assume. Do you think we should add one? It might be helpful for users that fail to download the java jar but then manually add it later.

dustinbyrne:
That's true, the addition of `appmap.jar` isn't reactive as it's currently only rendering the result of the initial download. I'm thinking we can queue that up as a future improvement for now as the number of user who will see the error state to begin with should be relatively low.

ahtrotta (approved):
Cool, I agree that we can make a follow-up improvement to be reactive to the `appmap.jar` file. :+1:

PR #785, opened 2023-08-30
Index receives api key

### LINKED ISSUE #784: Pass the API key when `index` is run from VS Code

ahtrotta:
Fixes #784 

This PR makes it so that when `index --watch` or `scan --watch` is executed from the VS Code extension, the `APPMAP_API_KEY` is set as an environment variable.


kgilpin (commented):
We are going to hold off on this for now since we don't have the app UI to utilize the data, and we want to focus on some other UX improvements that are more directly user-impacting.

kgilpin:
Deferred for now to focus on UX improvements.

PR #788, opened 2023-08-30
Require all users to sign in

### LINKED ISSUE #786: Require VS Code users to sign-in

ahtrotta:
Fixes #786 

This makes it so that any user that is logged out will see the sign-in page in the sidebar.


kgilpin:
Deferred for now to focus on UX improvements.

PR #794, opened 2023-09-05
feat: Display *.diff.sequence.json files

kgilpin:
File association for '*.diff.sequence.json'.

Related: https://github.com/getappmap/appmap-js/pull/1339

Screen Shot 2023-09-05 at 2 07 17 PM

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -20,6 +20,8 @@ import {
   spawn,
   verifyCommandOutput,
 } from '../services/nodeDependencyProcess';
+import { basename } from 'path';
+import { readFile } from 'fs/promises';
 
 export type FindingInfo = ResolvedFinding & {
   stackLocations?: StackLocation[];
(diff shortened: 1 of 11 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -103,31 +105,70 @@ export default class AppMapEditorProvider
...
+    }
+
+    if (appMapOrSequenceDiagramDiffUri.fsPath.endsWith('.diff.sequence.json')) {
+      sequenceDiagramData = await readFile(appMapOrSequenceDiagramDiffUri.fsPath, 'utf-8');
+      const appMapTokens = appMapOrSequenceDiagramDiffUri.fsPath.split('/');
+      const diffIndex = appMapTokens.lastIndexOf('diff');
+      if (diffIndex === -1)
+        return abortSequenceDiagramDiff(
+          `Expected to find a 'diff' directory in the path ${appMapOrSequenceDiagramDiffUri.fsPath}`
+        );
```
dustinbyrne, line 125: This violates the `CustomReadonlyEditorProvider` interface. What happens when `undefined` is returned? It'd be better to raise an exception otherwise this may result in unexpected behavior in the future.

dustinbyrne (changes requested):

kgilpin:
Tests passed on GitHub Actions.

PR #797, opened 2023-09-07
Clicking sign-in button cancels previous sign-in attempt

### LINKED ISSUE #792: Sign in button unresponsive after canceling URI prompt
A user clicks 'Cancel' instinctively after signing in on the VS Code prompt that asks 'Are you sure you want to open this URI?'. The user tried clicking 'sign in' again, but the button does nothing until the progress window ("Signing in with AppMap...") is closed. The user didn't see the progress window.

ahtrotta:
Fixes #792 

Previously, when signing in, if a user got this menu:

image

And then instinctively hit `Cancel` (as happened on a customer call), then they would get this progress menu in the bottom right corner:

image

If a user then hits the `Sign In` button again, nothing will happen as long as that progress menu is open. They have to hit the `Cancel` button on the progress menu before they can attempt to sign in again. This is a problem because it's easy to not notice the progress menu.

This PR makes it so that clicking the `Sign In` button automatically cancels any previous sign-in attempts and re-starts the sign-in process.

## QA Instructions

1. If you are signed in with AppMap Server, please sign out using the accounts button in the bottom left corner:
 
image

2. Click on the big blue Sign In button and then click Allow:

image

3. Click `Cancel`:

image

4. Then click the Sign In button again (it occasionally takes two clicks) and you should see this notification again:

image

### DIFF
--- a/src/authentication/appmapServerAuthenticationProvider.ts
+++ b/src/authentication/appmapServerAuthenticationProvider.ts
@@ -14,12 +14,19 @@ const debug = debuglog('appmap-vscode:AppMapServerAuthenticationProvider');
 
 const APPMAP_SERVER_SESSION_KEY = 'appmap.server.session';
 
+enum AuthFailure {
+  NotAuthorized = 'NotAuthorized',
+  UserCanceled = 'UserCanceled',
+  SignInAttempt = 'SignInAttempt',
+}
+
 export default class AppMapServerAuthenticationProvider implements vscode.AuthenticationProvider {
   private _onDidChangeSessions =
     new vscode.EventEmitter<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent>();
   readonly onDidChangeSessions = this._onDidChangeSessions.event;
 
   private session?: vscode.AuthenticationSession;
+  public customCancellationToken = new vscode.CancellationTokenSource();
 
   static enroll(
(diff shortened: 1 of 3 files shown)


src/authentication/appmapServerAuthenticationProvider.ts
```diff
@@ -128,19 +138,28 @@ export default class AppMapServerAuthenticationProvider implements vscode.Authen
...
+              dispose(disposables);
+              resolve(AuthFailure.SignInAttempt);
             });
           });
         }
       );
 
-      if (session) return session;
-      if (index === authnStrategies.length - 1) return undefined;
+      if (!Object.values(AuthFailure).includes(session as number))
```
dustinbyrne, line 158: These `AuthFailure` values would likely be better off `reject`ed rather than mixed in with the session type. At least that'd eliminate the need to access the enum as an object and cast the session to a number. TypeScript enums are just plain weird.

dustinbyrne (approved):
Looks good. I left a somewhat stylistic comment below but I'll leave it to your discretion.

src/authentication/appmapServerAuthenticationProvider.ts
ahtrotta: I made it so that the enum uses strings and the `session` can be cast as `AuthFailure` so it's a lot less weird.

PR #808, opened 2023-09-13
fix: Instructions page should be reactive to the Java agent file

### LINKED ISSUE #764: Instructions page should be reactive to the presence or absence of the Java agent file (`appmap.jar`)
Currently, the instructions page checks for the presence of `~/.appmap/lib/java/appmap.jar` just once, so it's not reactive to addition/deletion of that file. 

For example, if I:
1. Open a project while `appmap.jar` is present
2. Open the instructions page
3. Delete the `~/.appmap` folder

Then the instructions page still says that I have `~/.appmap/lib/java/appmap.jar` on my system:

image

We would like to make it reactive, so we'll probably need to add a file watcher.

zermelo-wisen:
Fixes #764

### DIFF
--- a/src/services/javaAssets.ts
+++ b/src/services/javaAssets.ts
@@ -7,6 +7,7 @@ import JavaAssetDownloader, { ProgressReporter } from '../lib/javaAssetDownloade
 import LockfileSynchronizer from '../lib/lockfileSynchronizer';
 import { fileExists } from '../util';
 import { promises as fs } from 'fs';
+import { clearInterval } from 'timers';
 
 export enum AssetStatus {
   Pending,
@@ -85,6 +86,7 @@ export default class JavaAssets {
   private static _status = AssetStatus.Pending;
   private static _onStatusChanged = new vscode.EventEmitter<AssetStatus>();
   private static outputChannel = vscode.window.createOutputChannel('AppMap: Assets');
+  private static intervalHandle?: NodeJS.Timer;
   public static onStatusChanged: vscode.Event<AssetStatus> = this._onStatusChanged.event;
 
   static get status() {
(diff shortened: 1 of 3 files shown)


src/services/javaAssets.ts
```diff
@@ -135,11 +137,47 @@ export default class JavaAssets {
...
+
+  static startAssetFileWatch(): void {
+    // We set up two file system watchers to monitor changes in the .appmap directory
+    // and its subdirectories. These watchers help us detect changes that may affect
+    // the result of the fileExists('~/.appmap/lib/java/appmap.jar') function.
+
+    // Watcher 1: wholePathWatcher
+    // This watcher monitors changes in the subdirectories [lib, java, appmap.jar] but
+    // does not detect changes in the .appmap root folder itself.
+    const wholePathWatcher = vscode.workspace.createFileSystemWatcher(
```
kgilpin, line 153: In my experience, any time we have a watcher we should also do a less frequent periodic scan of the directory (say every 10 seconds) - because watch events are unreliable especially for deletions. We want to make sure that our data is eventually consistent with the file system.

src/services/javaAssets.ts
```diff
@@ -135,11 +137,47 @@ export default class JavaAssets {
...
+      new vscode.RelativePattern(path.join(os.homedir(), '.appmap'), '**')
+    );
```
kgilpin, line 155: Can we be more specific with the watch pattern? .appmap can eventually contain a lot of files. We are only interested for this purpose in the lib folder. Right?

kgilpin: So what I’m saying is, constrain the pattern to the specific folders you care about because there may be other folders that contain many files.

kgilpin: Or just periodically check if the file you care about exists. This will be more robust than a watcher and have negligible performance implications for the user. Watchers just aren’t reliable.

zermelo-wisen: Thank you for pointing this out. Sorry, didn't know that there can be many files there. 

If I understand right (and based on my tests), to be specific, I need to create several watchers to cover each part of the path, since we want to make sure the exact path of `~/.appmap/lib/java/appmap.jar` is valid:

-watcher 1 `[~, .appmap]`
-watcher 2 `[.appmap, lib]`
-watcher 3 `[lib, java]`
-watcher 4 `[java, appmap.jar]`

If I omit, for example the first one, we can't react if the user deletes the `~/.appmap` folder.

I'm going to check to be sure if a single watcher like `[~/.appmap, lib/java/appmap.jar]` will do the job.

zermelo-wisen: > Or just periodically check if the file you care about exists. This will be more robust than a watcher and have negligible performance implications for the user. Watchers just aren’t reliable.

Ok, I just read your comment above. I'm going to convert it to a periodic (2 seconds?) check.

kgilpin: Ok just periodically list the directory `~/.appmap/lib/java` and that should tell you what you need to know. 2 seconds sounds good to me.

zermelo-wisen: Done.

src/services/javaAssets.ts
```diff
@@ -135,11 +137,25 @@ export default class JavaAssets {
     });
   }
 
-  static showOutput() {
+  static async onAssetFileCheckTick() {
+    const exists = await JavaAssets.assetsExist();
+    JavaAssets.status = exists ? AssetStatus.UpToDate : AssetStatus.Error;
```
dustinbyrne, line 142: Check out the definition of `installLatestJavaJar` above. If the status is `Updating`, we don't necessarily want to flag this as an error immediately; we have more information than that. Maybe it'd be best to exit early from the tick if we're in an `Updating` state.

dustinbyrne (changes requested):

src/services/javaAssets.ts
zermelo-wisen: Done

PR #815, opened 2023-09-15
feat: Rename AppMap auth provider to 'AppMap'

kgilpin:
`AppMap Server` is misleading, because it's really an account management provider and not an "API Server".

Rename the `AppMap Server` auth provider to `AppMap`.


brikelly (approved):
LGTM

PR #820, opened 2023-09-17
fix: Completion of the 'runtime analysis' step must come last

dustinbyrne:
The runtime analysis step no longer completes automatically upon opening the page if the user has not yet:
- Installed the agent,
- Recorded AppMaps, or
- Opened an AppMap

The previous steps must be completed prior to completing the final step.

Related to https://github.com/getappmap/appmap-js/issues/1383

### DIFF
--- a/test/system/tests/instructions.test.ts
+++ b/test/system/tests/instructions.test.ts
@@ -54,17 +54,26 @@ describe('Instructions tree view', function () {
       InstructionStepStatus.Complete
     );
 
-    await project.restoreFiles('**/appmap-findings.json');
+    // Assert that viewing the runtime analysis step prior to completing the
+    // previous steps does not mark the runtime analysis step as complete.
     await driver.appMap.openInstruction(InstructionStep.InvestigateFindings);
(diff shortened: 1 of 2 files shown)


test/system/tests/instructions.test.ts
```diff
@@ -54,20 +54,29 @@ describe('Instructions tree view', function () {
...
       InstructionStepStatus.Complete
     );
 
+    // All previous steps are now complete. Viewing the runtime analysis step
+    // should mark it as complete.
+    await driver.appMap.openInstruction(InstructionStep.InvestigateFindings);
+    await driver.appMap.assertInstructionStepStatus(
+      InstructionStep.InvestigateFindings,
+      InstructionStepStatus.Complete
+    );
```
dustinbyrne, line 78: Sorry to introduce additional logic to the system tests.  Importing the `InstructionsWebview` or `InstructionsTreeView` has proved difficult due to a downstream import of an `svg` file. `tsc` won't handle it. This solution side-steps the issue, but given the scope of the fix I don't think it's worth any more effort.

PR #821, opened 2023-09-17
fix: Always show 'no projects found' when viewing instructions if no project is open

dustinbyrne:
Resolves https://github.com/getappmap/appmap-js/issues/1386

To reproduce:
- `Ctrl+K F` will close the current workspace/folder
- Open any instructions page

### DIFF
--- a/test/system/tests/instructions.test.ts
+++ b/test/system/tests/instructions.test.ts
@@ -182,4 +182,32 @@ describe('Instructions tree view', function () {
     await driver.instructionsWebview.clickButton('Open the PROBLEMS tab');
     await driver.panel.problems.waitFor({ state: 'visible' });
   });
+
+  it('always opens the project picker in an empty/undefined workspace', async function () {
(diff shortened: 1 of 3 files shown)


test/system/tests/instructions.test.ts
```diff
@@ -182,4 +182,32 @@ describe('Instructions tree view', function () {
...
+      InstructionStep.OpenAppMaps,
+      InstructionStep.InvestigateFindings,
+    ];
+
+    for (let i = 0; i < steps.length; i++) {
+      const step = steps[i];
+      await driver.appMap.openInstruction(step);
+      await expectProjectPicker();
+    }
+  });
```
dustinbyrne, line 212: Similar to #820, I've added a system test for this behavior due to the fact that the import chain of `installGuideWebview` contains SVG files which aren't handled by `tsc`.

ahtrotta (approved):
Looks good and it worked well when I tested it manually

PR #823, opened 2023-09-18
feat: New users see a notification box when they create their first map

### LINKED ISSUE #817: New users see a notification box when they create their first map
Similar to how we launch this notification when a user opens a mappable project:
Image

We need to show a similar type of notification message when a user creates their very first map after installing the plugin for the first time, and only show it that one time. The purpose is to help guide new users towards their AppMaps, because they might not notice that they've been created or know how to even open them.

The text of the notification should be:
"You've created your first AppMap! Congratulations."
There should be one button as a primary action, colored in blue as above, with the text "Open your new AppMap"

This notification should only be seen at most once on a user's system.

zermelo-wisen:
Fixes #817.

### DIFF
--- a/test/integration/lib/firstAppMapNotification.test.ts
+++ b/test/integration/lib/firstAppMapNotification.test.ts
@@ -1,0 +1,37 @@
+import assert from 'assert';
+import * as vscode from 'vscode';
+import { SinonSandbox, createSandbox } from 'sinon';
+import ExtensionState from '../../../src/configuration/extensionState';
+import MockExtensionContext from '../../mocks/mockExtensionContext';
+import checkAndTriggerFirstAppMapNotification from '../../../src/lib/firstAppMapNotification';
+
+describe('First AppMap notification', () => {
+  let sinon: SinonSandbox;
+  let context: MockExtensionContext;
+  let extensionState: ExtensionState;
+
+  beforeEach(() => {
+    sinon = createSandbox();
+    context = new MockExtensionContext();
+    //appmapWatcher = new MockFileSystemWatcher();
+  });
+
+  afterEach(() => {
+    context.dispose();
+    sinon.restore();
+  });
+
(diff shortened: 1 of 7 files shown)


test/integration/lib/firstAppMapNotification.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+describe('First AppMap notification', () => {
+  let sinon: SinonSandbox;
+  let context: MockExtensionContext;
+  let extensionState: ExtensionState;
+
+  beforeEach(() => {
+    sinon = createSandbox();
+    context = new MockExtensionContext();
+    //appmapWatcher = new MockFileSystemWatcher();
```
ahtrotta, line 16: Please delete this comment

ahtrotta (changes requested):
This is cool! It's a little bit choppy, though. When I click the button, it first focuses on the `Code Objects` tree for a split second:

image

And then it focuses on the `AppMaps` tree, and it doesn't render all that well:

image

Can you change it so that the `Code Objects` tree does not get focused before the `AppMaps` tree?

Also, can you see if you can make the `AppMaps` tree open so that it occupies ~50% of the side panel and the top of the tree is always visible? Like this:

image

It might not be possible, but it's worth looking into.

zermelo-wisen:
I was able to prevent automatic focus on Code Objects. Now we focus directly on AppMaps tree.

However, I couldn't find a way to specify or cause a change in TreeView size. It seems that this is a related discussion: https://github.com/microsoft/vscode/issues/123715.

By the way, GitHub diff viewer in the Files changed tab here does not show the diffs well for the `installGuideWebView.ts` file. I guess it's because of the indentation change after introducing a new parameter which broke the line to make it "prettier". I suggest to check the diffs of this file inside VS Code.

ahtrotta (approved):
Great, thanks!

PR #829, opened 2023-10-02
Runtime Analysis is determined complete on multiple checks.

moxw:
Fixes: [#1403](https://github.com/getappmap/appmap-js/issues/1403)

feat(projectStateService.ts): enhance hasInvestigatedFindings check
test(instructions.test.ts): add test for runtime analysis status transitions

### DIFF
--- a/src/configuration/extensionState.ts
+++ b/src/configuration/extensionState.ts
@@ -13,6 +13,7 @@ export const Keys = {
     CONFIGURED_AGENT: 'appmap.applandinc.agentConfigured',
     RECORDED_APPMAP: 'appmap.applandinc.recordedAppMap',
     OPENED_APPMAP: 'appmap.applandinc.workspaces_opened_appmap',
+    OPENED_ANALYSIS: 'appmap.applandinc.workspaces_opened_analysis',
     FINDINGS_INVESTIGATED: 'appmap.applandinc.findingsInvestigated',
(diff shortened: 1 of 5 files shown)


ahtrotta (changes requested):
Apologies for not letting you know earlier, but we don't want to add system tests because they seem to be inherently flaky and fail intermittently. Can you make the system tests into integration tests or unit tests? Thanks!

moxw:
> Apologies for not letting you know earlier, but we don't want to add system tests because they seem to be inherently flaky and fail intermittently. Can you make the system tests into integration tests or unit tests? Thanks!

Added integration tests, have used `repeatUntil` otherwise it used to fail. Let me know if that's alright.

ahtrotta:
Thanks for converting the system test into an integration test. Using `repeatUntil` is fine.

I just tested this out, and if I generate maps, but don't complete the `Explore AppMaps` step, then I click on the `Runtime Analysis` step it shows a confusing message that setup is complete for the project:

image

Can you make it so that the `Runtime analysis` step is only marked as completed if all of the other instruction steps are completed? Thanks!

ahtrotta (dismissed):
Excellent, thanks @moxw! The code looks good and I tested this change and it works well.

ahtrotta (changes requested):
It looks like your changes caused one of the system tests to fail. Can you fix that? Thanks!

moxw:
> It looks like your changes caused one of the system tests to fail. Can you fix that? Thanks!

Fixed.

PR #830, opened 2023-10-05
feat: The Code Objects view also lists External Service Calls

zermelo-wisen:
Fixes https://github.com/getappmap/appmap-js/issues/1267.

### DIFF
--- a/src/lib/buildClassMap.ts
+++ b/src/lib/buildClassMap.ts
@@ -2,7 +2,12 @@ import { readFile } from 'fs';
 import { promisify } from 'util';
 import * as vscode from 'vscode';
 import * as path from 'path';
-import { ClassMapEntry, CodeObjectEntry, CodeObjectEntryRootType } from './CodeObjectEntry';
+import {
+  ClassMapEntry,
+  CodeObjectEntry,
+  CodeObjectEntryChildType,
+  CodeObjectEntryRootType,
+} from './CodeObjectEntry';
 
 const ROOT_ID = '<root>';
 
@@ -79,6 +84,16 @@ export async function buildClassMap(
         children: [normalizeClassMapEntry(cme)],
         labels: [],
       } as ClassMapEntry;
+    } else if (cme.type == CodeObjectEntryChildType.EXTERNAL_SERVICE) {
+      result = {
+        folder: cme.folder,
+        type: CodeObjectEntryRootType.FOLDER,
+        name: 'External Services',
+        static: false,
+        location: cme.location,
+        children: [normalizeClassMapEntry(cme)],
+        labels: [],
+      } as ClassMapEntry;
     }
     return result;
   }
@@ -134,8 +149,12 @@ export async function buildClassMap(
   // proper code object is missing from the classMap. They will appear as new root elements,
   // which is confusing in the UI. We just drop them here.
   const classMap = root.children
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-    .filter((child) => Object.values(CodeObjectEntryRootType).includes(child.type as any))
+    .filter(
+      (child) =>
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        Object.values(CodeObjectEntryRootType).includes(child.type as any) ||
+        child.type === CodeObjectEntryChildType.EXTERNAL_SERVICE
+    )
     .map((child) => {
       child.finalize();
       return child;
--- a/src/tree/classMapTreeDataProvider.ts
+++ b/src/tree/classMapTreeDataProvider.ts
@@ -99,15 +99,18 @@ export class ClassMapTreeDataProvider implements vscode.TreeDataProvider<vscode.
           : vscode.TreeItemCollapsibleState.None,
     } as CodeObjectTreeItem;
 
+    const isExternalRoute = codeObject.type.toString() === 'external-route';
     if (
       codeObject.type === CodeObjectEntryChildType.CLASS ||
       codeObject.type === CodeObjectEntryRootType.PACKAGE ||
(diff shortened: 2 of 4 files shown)


ahtrotta (changes requested):
Thanks @zermelo-wisen! There is a small change I'd like to see. When I click into the External Services tree, it shows a list of services that can be expanded:

image

But when I click on one of them, instead of expanding, it opens an appmap unless I specifically click on the `>` symbol. This is unexpected behavior, since the other expandable tree items do not behave this way. Can you make it so that clicking on an expandable child item like `api.openai.com` will not open an appmap and will instead expand the item and show its children?

zermelo-wisen:
> Thanks @zermelo-wisen! There is a small change I'd like to see. When I click into the External Services tree, it shows a list of services that can be expanded:
> 
> image
> 
> But when I click on one of them, instead of expanding, it opens an appmap unless I specifically click on the `>` symbol. This is unexpected behavior, since the other expandable tree items do not behave this way. Can you make it so that clicking on an expandable child item like `api.openai.com` will not open an appmap and will instead expand the item and show its children?

Done.

ahtrotta:
Hi @zermelo-wisen, thanks for the quick fix. The problem now is that when I click on the leaf node of the tree (for example, `POST http://127.0.0.1:9515/session/...` in the tree in the image above) it doesn't open the appmap. Apologies if I was unclear before, but here's what we need:

- [ ] Clicking on an expandable tree member (for example, `127.0.0.1:9515` or `api.openai.com` in the image above) should expand the tree member and show its children.
- [ ] Clicking on a leaf node in the tree (for example, `POST http://127.0.0.1:9515/session/...` in the image above) should open the appmap with the code object selected.

zermelo-wisen:
> Hi @zermelo-wisen, thanks for the quick fix. The problem now is that when I click on the leaf node of the tree (for example, `POST http://127.0.0.1:9515/session/...` in the tree in the image above) it doesn't open the appmap. Apologies if I was unclear before, but here's what we need:
> 
> * [ ]  Clicking on an expandable tree member (for example, `127.0.0.1:9515` or `api.openai.com` in the image above) should expand the tree member and show its children.
> * [ ]  Clicking on a leaf node in the tree (for example, `POST http://127.0.0.1:9515/session/...` in the image above) should open the appmap with the code object selected.

Ok. In this example `type` of `127.0.0.1:9515` is `external-service` and `type` of `POST http://127.0.0.1:9515/session/...` is  `external-route`. Executing `appmap.openCodeObjectInAppMap` command with `fqId` of `external-route` item opens the app map but does not focus on the call itself. If I give its parent's `fqid` to the command, it focuses on the item representing its parent, which is `external-service` `127.0.0.1:9515` for example. If this behaviour is ok, I will do it this way.

`external-route` type is not a member in `CodeObjectEntryChildType` enum. Should I add it there? Should I add it to InspectableTypes, too?

By the way, I forgot to mention in my previous message that we introduced the dual behaviour (`>` expands, _click_ opens appmap) on items having package or class type, in a previous requirement/issue.

ahtrotta:
> Ok. In this example `type` of `127.0.0.1:9515` is `external-service` and `type` of `POST http://127.0.0.1:9515/session/...` is `external-route`. Executing `appmap.openCodeObjectInAppMap` command with `fqId` of `external-route` item opens the app map but does not focus on the call itself. If I give its parent's `fqid` to the command, it focuses on the item representing its parent, which is `external-service` `127.0.0.1:9515` for example. If this behaviour is ok, I will do it this way.
> 
> `external-route` type is not a member in `CodeObjectEntryChildType` enum. Should I add it there? Should I add it to InspectableTypes, too?
> 
> By the way, I forgot to mention in my previous message that we introduced the dual behaviour (`>` expands, _click_ opens appmap) on items having `package` or `class` type, in a previous requirement/issue.

Ah, okay I didn't realize that the dual behavior was expected. 

- [ ] Please revert this PR to the way it was originally with the dual behavior. 
- [ ] If the `type` is `external-service` you can pass the parent `fqid`. 
- [ ] Also, don't add `external-route` to `CodeObjectEntryChildType`.

Thanks!

zermelo-wisen:
> > Ok. In this example `type` of `127.0.0.1:9515` is `external-service` and `type` of `POST http://127.0.0.1:9515/session/...` is `external-route`. Executing `appmap.openCodeObjectInAppMap` command with `fqId` of `external-route` item opens the app map but does not focus on the call itself. If I give its parent's `fqid` to the command, it focuses on the item representing its parent, which is `external-service` `127.0.0.1:9515` for example. If this behaviour is ok, I will do it this way.
> > `external-route` type is not a member in `CodeObjectEntryChildType` enum. Should I add it there? Should I add it to InspectableTypes, too?
> > By the way, I forgot to mention in my previous message that we introduced the dual behaviour (`>` expands, _click_ opens appmap) on items having `package` or `class` type, in a previous requirement/issue.
> 
> Ah, okay I didn't realize that the dual behavior was expected.
> 
> * [ ]  Please revert this PR to the way it was originally with the dual behavior.
> * [ ]  If the `type` is `external-service` you can pass the parent `fqid`.
> * [ ]  Also, don't add `external-route` to `CodeObjectEntryChildType`.
> 
> Thanks!

I think you wanted to say, "If the type is external-route you can pass the parent fqid.", which means both parent ('external-service') and child ('external-route') will be clickable. Child click will be exactly like the parent click because we will use parent fqid when clicking the child.

ahtrotta:
> I think you wanted to say, "If the type is external-route you can pass the parent fqid.", which means both parent ('external-service') and child ('external-route') will be clickable. Child click will be exactly like the parent click because we will use parent fqid when clicking the child.

Yes, exactly, thank you :+1:

ahtrotta (approved):
Great, thanks!

Issue #833, opened 2023-10-10
Using the "Delete AppMaps" context menu does not work


brikelly:
Right click on a map category, and choose the `AppMap View: Delete AppMaps` option. This currently does not work at all, but it should delete the maps in that folder/category.

![Image](https://github.com/getappmap/vscode-appland/assets/573108/8c357092-ce76-43fd-9aef-7f29c4d28466)

It can also be selected at the top level, which would be the same as deleting all maps in the project.

kgilpin:
@dividedmind @dustinbyrne  does this collide with what the indexer is trying to do on AppMap deletion?

dividedmind:
> @dividedmind @dustinbyrne does this collide with what the indexer is trying to do on AppMap deletion?

No, both use `force` so the rm will succeed even when the file already doesn't exist.

PR #834, opened 2023-10-11
Emit delete appmap event reliably

moxw:
Fixes: #833 

feat(deleteAppMap.ts): ensure reliable delete event emission by adding retry on metadata.json removal

### DIFF
--- a/src/lib/deleteAppMap.ts
+++ b/src/lib/deleteAppMap.ts
@@ -20,6 +20,9 @@ export default async function deleteAppMap(
 
   const indexDir = uri.fsPath.substring(0, uri.fsPath.lastIndexOf('.appmap.json'));
 
+  // This triggers the FileSystemWatcher to emit a delete event for the AppMap file reliably.
+  await retry(async () => rm(`${indexDir}/metadata.json`, { force: true }));
+
   await retry(async () => rm(uri.fsPath, { force: true }));
   await retry(async () => {
     for (const file of await promisify(glob)(`${indexDir}/*`)) {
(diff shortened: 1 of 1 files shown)


src/lib/deleteAppMap.ts
```diff
@@ -20,6 +20,9 @@ export default async function deleteAppMap(
 
   const indexDir = uri.fsPath.substring(0, uri.fsPath.lastIndexOf('.appmap.json'));
 
+  // This triggers the FileSystemWatcher to emit a delete event for the AppMap file reliably.
+  await retry(async () => rm(`${indexDir}/metadata.json`, { force: true }));
```
dustinbyrne, line 24: This should be handled just below during the removal of `indexDir/*`. Is this causing a different behavior?

moxw: Yes, it appears the general deletion wasn't triggering VSCode's delete event, necessitating this specific file removal.

dustinbyrne: It's difficult for me to see how this change affects anything. Take a look at [`IndexJanitor`](https://github.com/getappmap/vscode-appland/blob/293316a9bac11d0c74733f4a19a44184080512df/src/lib/indexJanitor.ts) for more details.

moxw: @dustinbyrne, I found that deleting metadata.json specifically triggers the VSCode delete event more reliably than removing indexDir/*. It'd be great if you could test this on your end too.

ahtrotta: @dustinbyrne for whatever reason the change that @moxw made does indeed fix the problem :man_shrugging:

ahtrotta (approved):

PR #842, opened 2023-11-10
fix: Improve resource usage in file watchers and finders

### LINKED ISSUE #839: High CPU usage by file watchers
We create a couple of recursive filesystem watchers causing high CPU usage; with some projects it makes the system almost unusable. VSCode documentation contains suggestions on how to approach filesystem watchers to limit resource use.

dividedmind:
- Don't create per-folder watchers when not necessary. VSCode by default watches all folders.
- Use files.watcherExcludes to exclude locations from initial search. This avoids descending into node_modules and similar locations.

Fixes #839


dustinbyrne:
FWIW I'm still seeing a ton of CPU utilization when loading the extension in `appmap-node`.
image

Though, it is down one ripgrep process from what it was prior to this change
image

dustinbyrne (approved):
Looks good! Performance has definitely improved without the use of `ripgrep` on startup.

PR #845, opened 2023-11-21
feat: Search + Explain with Navie

kgilpin:


src/webviews/getWebviewContent.ts
```diff
@@ -28,12 +29,6 @@ export default function getWebviewContent(
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta http-equiv="Content-Security-Policy" content="
```
kgilpin, line 31: @dividedmind as our resident security person, and I think you added this initially, why was this added? We need to allow the UI to invoke JSON-RPC to our services.

dividedmind: This is to prevent the webview content from being coerced to load external resources, which could cause data leak or worse. I don't think it's a good idea to remove this altogether; instead a directive like `connect-src http://localhost:12345;` should be added. Although ideally I think the UI shouldn't talk to the services directly at all; instead the extension should intermediate.

kgilpin:
See https://github.com/getappmap/vscode-appland/pull/859

PR #849, opened 2023-11-29
Fix source links with drive letters on Windows

### LINKED ISSUE #840: Some source code links don't work in Windows
Some source code links in the AppMap do not work correctly in Windows in VS Code. They work correctly in JetBrains. For example, clicking on this does nothing:

Image

ahtrotta:
Fixes #840

### DIFF
--- a/src/editor/appmapEditorProvider.ts
+++ b/src/editor/appmapEditorProvider.ts
@@ -1,6 +1,5 @@
 import * as vscode from 'vscode';
 import * as fs from 'fs';
-import { parse } from 'path';
 import { Telemetry, DEBUG_EXCEPTION } from '../telemetry';
 import { version } from '../../package.json';
 import ExtensionState from '../configuration/extensionState';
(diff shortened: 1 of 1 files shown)


src/editor/appmapEditorProvider.ts
```diff
@@ -489,7 +490,18 @@ export default class AppMapEditorProvider
...
+      const parsedLocation = parse(location);
+      const { root } = parsedLocation;
+
+      // Handle the situation where the location is a Windows file path with a drive letter
+      // For example, C:/Users/user/some/directory/file.rb:4
+      if (root.includes(':')) {
+        const driveLetter = tokens.shift();
+        tokens[0] = `${driveLetter}:${tokens[0]}`;
+      }
+
```
dividedmind, line 504: This seems needlessly complicated. How about `const [, path, line] = location.match(/^(.*?)(?::(\d+))?$/)`?

dividedmind (changes requested):

src/editor/appmapEditorProvider.ts
ahtrotta: Sure, I changed it :+1:

PR #855, opened 2024-01-02
Authenticate the chat search frontend

dustinbyrne:
This is necessary to make authenticated HTTP requests, such as providing message feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,7 +220,12 @@
         "appMap.applandUrl": {
           "type": "string",
           "default": "https://getappmap.com",
-          "description": "URL of an AppLand cloud instance for AppMap storage"
+          "description": "URL of AppMap"
+        },
+        "appMap.apiUrl": {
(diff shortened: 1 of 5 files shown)


kgilpin:
What’s message feedback?

Can we or should we do this through the RPC layer instead? That is more reusable with JetBrains.

dustinbyrne:
It's just a standard API call via browser `fetch` so it should also work in Jetbrain's JCEF. For that reason I don't know that it's necessarily worth routing it through RPC call.

kgilpin:
Can you explain what the feature is that this is supporting?

dustinbyrne:
This is used by the feedback buttons

[example](https://github.com/getappmap/vscode-appland/assets/8737782/9a9cb620-201f-472f-bbec-9e0f3bd65654)

kgilpin (approved):

PR #856, opened 2024-01-03
fix: Ensure a deleted AppMap is removed from the tree

kgilpin:
AppMap deletion does not reliably remove the AppMap from the tree view. Ensure that it does.


kgilpin:
FYI this is blocked by the "ReadableStream" problem.

PR #858, opened 2024-01-05
feat: Update VSCode engine to v1.82.0

dustinbyrne:
This is the first version of VSCode that ships Node 18.

This will affect approx. 10% of users who are running < v1.82.0.


dustinbyrne:
If merging, please squash

PR #859, opened 2024-01-10
feat: AppMap AI: Explain

kgilpin:
## Dependencies

- [x] Merge https://github.com/getappmap/appmap-js/pull/1528

## Commands

* `AppMap: AI Explain` top-level command

Running this command opens the Explain UI page.

Screen Shot 2024-01-10 at 7 31 44 PM

* `Explain with AppMap AI` quick search command from an open editor.

When the user selects some code in a supported language, the "light bulb" appears. This is known in VSCode as "Quick Search", it opens the AppMap Explain AI page with the question pre-populated with the user's selection.

Screen Shot 2024-01-10 at 7 32 15 PM

### AppMap state

AppMap state (basically, the filter popup), can be saved, applied, and transferred to and from the clipboard. This functionality should work for the AppMaps shown in the Explain UI as well.

### AppMap RPC server

The `appmap index` command should be launched with options `--port 0`. With this option, the index process will run a JSON-RPC service on an OS-assigned port. JSON-RPC was chosen for the transport because it's also used by the Language Server Protocol. 

The port is printed to stdout by the index command, so the Plugin should scan stdout to determine this port.

When an Explain UI page is launched, it should be provided with this port number via the `appmapRpcPort` property. 

### API key

The AppMap RPC server makes HTTP calls to the AppMap API services, including a hosted completion AI. Therefore, the user must have an AppMap API key to run the RPC server. The API key is provided to the `appmap index` command using the environment variable `APPMAP_API_KEY`. 

Not shown in this PR - the AppMap API key is also provided to the Help chat and Explain chat directly as a property, so that the chat UI can send user feedback to the AppMap API services (thumbs up / thumbs down).


src/webviews/getWebviewContent.ts
```diff
@@ -28,12 +29,6 @@ export default function getWebviewContent(
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta http-equiv="Content-Security-Policy" content="
-      default-src 'none';
-      img-src ${webview.cspSource} data:;
-      script-src ${webview.cspSource} 'unsafe-eval';
-      style-src ${webview.cspSource} 'unsafe-inline';
-    ">
```
dividedmind, line 36: We shouldn't remove the CSP; if we want the UI to be able to connect to the server (which I don't think it should be able to do directly) then we need to set it as small as possible, cf. https://github.com/getappmap/vscode-appland/pull/845#discussion_r1434673224

dividedmind: I added the CSP back in 3a01a83 — what was needed was `fetch-src http://localhost:${rpcPort}`.

src/services/indexProcessWatcher.ts
```diff
@@ -1,21 +1,55 @@
...
+    this.stdoutBuffer = this.stdoutBuffer.slice(-100);
+
+    const portStr = lines
+      .map((line) => {
+        const match = line.match(/^Running JSON-RPC server on port: (\d+)$/);
+        if (match) return match[1];
+      })
+      .find(Boolean);
+    if (portStr) {
+      vscode.window.showInformationMessage(`AppMap index process listening on port ${portStr}`);
```
dustinbyrne: I think it'd be worth moving this to the `AppMap: Services` output window instead of a popup before release.

dustinbyrne (approved):

PR #872, opened 2024-01-23
feat: Navie prompts the user to choose a directory

### LINKED ISSUE #869: Navie prompts the user to choose a directory
When there are multiple appmap.yml files available within a workspace, Navie should prompt the user to choose a directory.

The port number and associated directory should be logged for each Navie command.

kgilpin:
When there are multiple appmap.yml in a workspace, the user must be prompted to choose which one they want to use for their question. Each appmap.yml will correspond to a Navie RPC port, to which the question should be directed.

If there are no workspaces open with appmap.yml, then as before the Explain command won't run.

If there is exactly one index process with a ready RPC port, then this process is selected.

Otherwise, the user is prompted to choose which directory they want to run the Explain command in.

Screen Shot 2024-01-23 at 4 36 59 PM

Fixes #869

### DIFF
--- a/src/lib/selectIndexProcess.ts
+++ b/src/lib/selectIndexProcess.ts
@@ -1,0 +1,91 @@
+import * as vscode from 'vscode';
+import IndexProcessWatcher from '../services/indexProcessWatcher';
+import { NodeProcessService } from '../services/nodeProcessService';
+import { workspaceServices } from '../services/workspaceServices';
+
+export type IndexProcess = {
+  configFolder: string;
+  rpcPort: number;
+};
+
+export enum ReasonCode {
+  NoIndexProcessWatchers = 1,
+  NoReadyIndexProcessWatchers = 2,
+  NoSelectionMade = 3,
+}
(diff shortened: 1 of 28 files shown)


src/lib/selectIndexProcess.ts
```diff
@@ -0,0 +1,91 @@
...
+      (service) =>
+        service.processes.filter((process) => process.id === 'index') as IndexProcessWatcher[]
+    );
+
+  return indexProcessWatchers;
+}
+
+export default async function selectIndexProcess(
+  workspace?: vscode.WorkspaceFolder
+): Promise<IndexProcess | ReasonCode | undefined> {
```
dustinbyrne, line 44: I like this pattern. I think I might use it :+1:

dustinbyrne (approved):

src/lib/selectIndexProcess.ts
kgilpin: Thanks!

PR #873, opened 2024-01-25
feat: Drop the quickExplain code action (light bulb action)

### LINKED ISSUE #870: Drop the Navie light bulb tool

kgilpin:
Fixes #870


kgilpin:
Testers are using this. Should we wait to process their feedback before dropping it?

kgilpin:
I think this can be fixed and should not be closed. Dropping.

PR #909, opened 2024-03-08
feat: Navie works without AppMap

dustinbyrne:
This change will spawn an index process as an RPC server (no file watching) if no index processes are available. This means in workspaces where appmap.yml is not present, Navie can still be used.

TODO:
- [ ] Global RPC port isn't getting cleaned up (async issue?)
- [ ] Do we even need a global process if it's spawned per extension only upon using Navie?
- [ ] Tests

### DIFF
--- a/src/services/indexProcessWatcherGlobal.ts
+++ b/src/services/indexProcessWatcherGlobal.ts
@@ -1,0 +1,132 @@
+import * as vscode from 'vscode';
+import ExtensionSettings from '../configuration/extensionSettings';
+import { NodeProcessService } from './nodeProcessService';
+import { ProcessId, ProcessWatcher, ProcessWatcherOptions } from './processWatcher';
+import LockfileSynchronizer from '../lib/lockfileSynchronizer';
+import { ProgramName, getModulePath } from './nodeDependencyProcess';
+
+// This class represents a single, global instance of the AppMap index process watcher.
+// It is used in the absence of a workspace or AppMap configuration file.
+// It does not watch the file system; It is only used as an RPC server.
+export default class IndexProcessWatcherGlobal extends ProcessWatcher {
+  private _onRpcPortChange = new vscode.EventEmitter<number | undefined>();
+  private onRpcPortChange = this._onRpcPortChange.event;
+
+  stdoutBuffer = '';
+
+  constructor(
+    context: vscode.ExtensionContext,
+    modulePath: string,
+    cwd: string,
+    env?: NodeJS.ProcessEnv
+  ) {
+    const args = ['index', '--port', '0'];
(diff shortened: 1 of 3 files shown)


src/services/indexProcessWatcherGlobal.ts
```diff
@@ -0,0 +1,132 @@
+import * as vscode from 'vscode';
+import ExtensionSettings from '../configuration/extensionSettings';
+import { NodeProcessService } from './nodeProcessService';
+import { ProcessId, ProcessWatcher, ProcessWatcherOptions } from './processWatcher';
+import LockfileSynchronizer from '../lib/lockfileSynchronizer';
+import { ProgramName, getModulePath } from './nodeDependencyProcess';
+
+// This class represents a single, global instance of the AppMap index process watcher.
+// It is used in the absence of a workspace or AppMap configuration file.
+// It does not watch the file system; It is only used as an RPC server.
```
kgilpin, line 10: A problem is that without a working directory containing an appmap.yml, the RPC server can't serve the AppMap stats. So the left hand side panel of Navie will be empty and remain empty even when the user makes AppMaps.

src/services/indexProcessWatcherGlobal.ts
```diff
@@ -0,0 +1,132 @@
...
+export default class IndexProcessWatcherGlobal extends ProcessWatcher {
```
kgilpin, line 11: Having a global Navie seems like a necessary thing to do, so that Navie can always be available.

It might also have some really useful benefits going forward. For example, the user wouldn't have to choose a working directory; all directories containing appmap.yml and AppMaps would be considered for the Search phase.  And the user's question could span across multiple workspace folders that all have AppMaps. 

But to do this will take some re-engineering of the backend `index` command, we can't address it purely from the UI.

src/services/indexProcessWatcherGlobal.ts
dustinbyrne: For now this process will only be used if the workspace has no configuration. If they create one, a 'regular' index process will spin up and Navie will communicate with the process created for that workspace instead.

I suppose there's some extra work to be done to propagate the change in port in real-time to existing Navie frontends.

kgilpin: I would like to go with the design where the Navie process is always running and the configuration is updated via JSON RPC

See the draft PR in appmap-js

kgilpin: We don’t want to have Chat windows that are open but connected to RPC ports that are either closed or out of date. In the design of this proposed PR that will happen. Using the design of the appmap-js PR, it won’t.

PR #912, opened 2024-03-12
feat: Introduce the RPC background process

dustinbyrne:
This PR removes the `--port` flag from the indexer in favor of running a single RPC server process on demand.

The process will only start upon extension initialization, and spans all projects within the workspace.


src/services/rpcProcessService.ts
```diff
@@ -3,90 +3,89 @@ import RpcProcessWatcher from './rpcProcessWatcher';
...
-    this.portChangeListener = this.processWatcher.onRpcPortChange((port) => {
-      this.rpcPort = port;
-      this._onRpcPortChange.fire(port);
-    });
   }
 
-  public get available(): boolean {
-    return this.processWatcher?.running ?? false;
-  }
+  private syncAppMapConfigurations(): void {
```
kgilpin, line 52: Can we also do this on an Interval in case we miss a config change for some reason (eg race condition)?

PR #915, opened 2024-03-14
CLI installer fix

### LINKED ISSUE #913: Unknown arguments: ms-enable-electron-run-as-node, msEnableElectronRunAsNode when using UI Installer
When running the appmap installer from the VS Code window the installer CLI command will return the following error:

```
ELECTRON_RUN_AS_NODE=true /Applications/Visual\ Studio\ Code.app/Contents/Frameworks/Code\ Helper\ \(Plugin\).app/Contents/MacOS/Code\ Helper\ \(Plug
in\) /Users/petecheslock/Library/Application\ Support/Code/User/globalStorage/appland.appmap/node_modules/@appland/appmap/built/cli.js install --ms-enable-electron-run-as-node -d /User
s/petecheslock/repos/sample_rails_app
Code Helper (Plugin) install [directory]

Install and configure an AppMap language agent

Positionals:
  directory, d  Directory in which to install (deprecated; use -d)
                                                         [string] [default: "."]

Options:
      --version                  Show version number                   [boolean]
  -v, --verbose                  Run with verbose logging              [boolean]
      --help                     Show help                             [boolean]
  -p, --project-type             Specifies the project dependency manager.
                                 Case-insensitive.
                                 Supported project types: Bundler, Gradle,
                                 Maven, undefined, pip, pipenv, poetry  [string]
      --interactive              Whether to interact with the user (assuming
                                 there's a TTY).       [boolean] [default: true]
      --overwrite-appmap-config  Whether to overwrite the appmap.yml file.
                                                                       [boolean]
      --installer-name           Installer name to use, in case of ambiguity.
                                                                        [string]
      --build-file               Build file name to use, in case of ambiguity.
                                                                        [string]
Unknown arguments: ms-enable-electron-run-as-node, msEnableElectronRunAsNode
```

https://www.loom.com/share/b151eedb285847d28cb95d61bd18370f

ahtrotta:
Fixes #913 

Only add the `--ms-enable-electron-run-as-node` flag for versions of VS Code prior to 1.86.0.

### DIFF
--- a/src/commands/installAgent.ts
+++ b/src/commands/installAgent.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 import os from 'os';
 import { join } from 'path';
+import semver from 'semver';
 import { INSTALL_BUTTON_ERROR, Telemetry } from '../telemetry';
 import { NodeProcessService } from '../services/nodeProcessService';
 import { Installer } from './installer';
@@ -48,7 +49,13 @@ function electronCommand(globalStorageDir: string, installLocation: string): str
   // This is the exact behavior we want to emulate in the terminal, so we'll use it here as well.
   const nodePath = escapePath(process.execPath);
   const cliPath = join(globalStorageDir, 'node_modules', '@appland', 'appmap', 'built', 'cli.js');
-  const flags = ['--ms-enable-electron-run-as-node', '-d', installLocation];
+  const flags = ['-d', installLocation];
+
+  // This flag was required in VS Code versions prior to 1.86.0
+  const vsCodeVersion = semver.coerce(vscode.version);
+  if (vsCodeVersion && semver.lt(vsCodeVersion, '1.86.0'))
+    flags.push('--ms-enable-electron-run-as-node');
+
   return `ELECTRON_RUN_AS_NODE=true ${nodePath} ${cliPath} install ${flags.join(' ')}`;
 }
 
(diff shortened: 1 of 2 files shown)


src/commands/installAgent.ts
```diff
@@ -48,7 +48,7 @@ function electronCommand(globalStorageDir: string, installLocation: string): str
   // This is the exact behavior we want to emulate in the terminal, so we'll use it here as well.
   const nodePath = escapePath(process.execPath);
   const cliPath = join(globalStorageDir, 'node_modules', '@appland', 'appmap', 'built', 'cli.js');
-  const flags = ['--ms-enable-electron-run-as-node', '-d', installLocation];
```
dustinbyrne, line 51: The `--ms-enable-electron-run-as-node` flag used to be required but looks to have been removed at some point. Can you verify this happened before v1.82.0? That's the version of VS Code we require.

As long as it happened before, we can remove this altogether. Otherwise we'll need to check the VS Code version to see if we need to include the flag.

dustinbyrne (approved):

src/commands/installAgent.ts
ahtrotta: I checked and it appears that we need `--ms-electron-enable-run-as-node` for versions of VS Code prior to 1.86.0, and then that flag breaks things for version 1.86.0 or later. So I changed the code to detect the version of VS Code and add the flag accordingly.

PR #926, opened 2024-03-24
feat: Allow configuring CLI environment

dividedmind:
Also, if Azure OpenAI configuration is detected in the configured environment, use the OpenAI key from secrets storage for it.

obraz

### DIFF
--- a/src/services/processWatcher.ts
+++ b/src/services/processWatcher.ts
@@ -50,31 +50,20 @@ async function accessToken(): Promise<string | undefined> {
 export async function loadEnvironment(
   context: vscode.ExtensionContext
 ): Promise<NodeJS.ProcessEnv> {
-  const env: Record<string, string> = {};
-  const environmentVariables: { env: string; provider: LookupFunction }[] = [
-    {
-      env: 'APPMAP_API_URL',
-      provider: () => Promise.resolve(ExtensionSettings.apiUrl),
-    },
-    {
-      env: 'APPMAP_API_KEY',
-      provider: accessToken.bind(null),
-    },
-    {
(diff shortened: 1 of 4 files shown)


src/services/processWatcher.ts
```diff
@@ -50,32 +50,20 @@ async function accessToken(): Promise<string | undefined> {
...
-  ];
-
-  for (const { env: key, provider } of environmentVariables) {
-    const value = await provider();
-    if (value) env[key] = value;
+  const env: Record<string, string> = ExtensionSettings.appMapCommandLineEnvironment ?? {};
+  const openAIApiKey = await getOpenAIApiKey(context);
+  if (openAIApiKey) {
+    if ('AZURE_OPENAI_API_VERSION' in env) env.AZURE_OPENAI_API_KEY = openAIApiKey;
+    else env.OPENAI_API_KEY = openAIApiKey;
```
dustinbyrne, line 57: In my testing `OPENAI_API_KEY` isn't being set on `env`. Copying `ExtensionSettings.appMapCommandLineEnvironment` into another object does seem to fix this, though I'm not sure why. The type returned by `getConfiguration(..).get(..)` is a proxy, so maybe the return type is misleading and it doesn't allow modifications?

e.g.
```ts
  public static get appMapCommandLineEnvironment(): Record<string, string> | undefined {
    return {...vscode.workspace.getConfiguration('appMap').get('commandLineEnvironment') ?? {}};
  }
```

dividedmind: Ah, good catch, I believe you're right!

dividedmind: I added a fixup, can you double check? Thanks!

PR #928, opened 2024-03-25
Improve the resilience of asset fetching and use CLI binaries

### LINKED ISSUE #919: `yarn` is not respecting OS certificate store
Proxied traffic over SSL results in `unable to get local issuer certificate`

### LINKED ISSUE #925: The Java agent JAR file is downloaded from Maven Central
The plugin should first try to download the latest version of the agent JAR from Maven Central. If it cannot. it should fall back to the current behavior of downloading from GitHub.

Failures should be written to the log. But, because the plugin now bundles a version of the agent JAR, they should just be warnings.

### LINKED ISSUE #927: All platforms download and use the binary CLIs instead of node modules

dustinbyrne:
(wip)

- All platforms will now use the pre-built binaries
- All asset downloads will respect VS Code proxy settings (or `HTTP_PROXY`, `HTTPS_PROXY`)
- Java agent downloads will first source from Maven, falling back to GitHub Releases
- All asset downloads can define their own fallback locations for both locating the latest version and downloading the actual asset.

Resolves #919 (yarn is removed)
Resolves #925
Resolves #927


src/assets/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      }
+    }
+  }
+}
+
+class GitHubVersionResolver implements VersionResolver {
+  constructor(private readonly repo: string) {}
+
+  async getLatestVersion(): Promise<string | undefined> {
+    const res = await tryRequest(`https://api.github.com/repos/${this.repo}/releases/latest`);
```
kgilpin, line 57: Is it possible to avoid this entirely?

kgilpin: I guess what I mean is, if we can get the same information from NPM or Maven, those are better candidates because we don't see them being blocked (far as I can tell). Whereas blocking GitHub API seems common for some reason.

We could query api.getappmap.com for this information, since that site must always be available anyway?

dustinbyrne: I'm using this as the lowest priority fallback. It won't attempt this unless Maven is blocked for some reason.

dustinbyrne: It'd definitely be good to add some alternate strategies by mirroring this elsewhere. At the moment, the CLI binaries don't have any solid fallback options. I was thinking S3 would be another good option.

PR #933, opened 2024-04-03
Include .tsx files for node projects

ahtrotta:
This adds `.tsx` files for JavaScript projects.


kgilpin:
My only question is - are there any others?

ahtrotta:
> My only question is - are there any others?

As far as I can tell we should have all of our bases covered now.

dustinbyrne (approved):

PR #941, opened 2024-04-15
feat: Add the LLM configuration frontend

dustinbyrne:


src/services/rpcProcessService.ts
```diff
@@ -177,6 +179,25 @@ export default class RpcProcessService implements Disposable {
     return service;
   }
 
+  public async checkForEnvironmentVariableChanges(): Promise<void> {
```
ahtrotta, line 182: This isn't being called anywhere

ahtrotta (approved):

PR #942, opened 2024-04-16
Open Navie instead of instrutions for new users

ahtrotta:


kgilpin:
I don’t think this is what we want. 

We want the new Navie panel and we want to stop opening the install flow. 

But not open Navie.

PR #943, opened 2024-04-16
Do not automatically open the instructions

ahtrotta:
This PR makes it so that the AppMap Recording Instructions no longer open automatically for a new user. Also, the popup message prompting the user to create AppMap recordings no longer appears.


src/extension.ts
```diff
@@ -224,8 +222,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
     badge.initialize(projectStates);
     context.subscriptions.push(badge);
 
-    InstallGuideWebView.register(context, projectStates);
-    const openedInstallGuide = await vscode.commands.executeCommand('appmap.tryOpenInstallGuide');
+    InstallGuideWebView.register(context, projectStates, extensionState);
```
kgilpin, line 225: `extensionState` does not appear to be needed here.

kgilpin:
We should figure out if this functionality is going to remain: ```  const msg: Array<string> = [];
  if (numProjects === 1) {
    msg.push('AppMap is ready to map and analyze this project.');
  } else {
    // Must be greater than one, otherwise we'd already have returned
    if (numProjects === numInstallable) {
      msg.push('AppMap is ready to map and analyze every project within this workspace.');
    } else {
      msg.push(`AppMap is ready to map and analyze ${numInstallable} projects in this workspace.`);
    }
  }
  msg.push('Open the setup instructions?');
```

File: src/extension.ts [hunk 2]
```diff
@@ -282,7 +279,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<AppMap
     downloadLatestJavaJar(context);
     getAppmapDir(context, workspaceServices);
 
-    if (!openedInstallGuide && !SignInManager.shouldShowSignIn())
+    if (!extensionState.hasViewedInstallGuide && !SignInManager.shouldShowSignIn())
```
kgilpin, line 282: So this would now never happen if the user had viewed the install guide once before?

kgilpin (approved):

PR #947, opened 2024-04-19
feat: Send active code editor selection to Navie

kgilpin:
Propagate code selection to Navie. 

## Example

Select code:

Screen Shot 2024-04-19 at 1 43 41 PM

Open Navie, the code selection is populated the same way as if you'd used the light bulb:

Screen Shot 2024-04-19 at 1 44 09 PM

- [x] Test latest appmaps event
- [x] Pass AppMaps as the event argument

### DIFF
--- a/resources/versions.json
+++ b/resources/versions.json
@@ -1,5 +1,5 @@
 {
-  "appmap": "3.135.1",
+  "appmap": "3.136.0",
   "scanner": "1.86.0",
-  "appmap-java.jar": "1.26.5"
+  "appmap-java.jar": "1.26.7"
 }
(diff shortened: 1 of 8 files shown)


resources/versions.json
```diff
@@ -1,5 +1,5 @@
 {
-  "appmap": "3.135.1",
+  "appmap": "3.136.0",
   "scanner": "1.86.0",
-  "appmap-java.jar": "1.26.5"
+  "appmap-java.jar": "1.26.7"
```
dustinbyrne, line 4: FWIW manual updates to this file shouldn't be necessary as it'll synchronize automatically before publish
https://github.com/getappmap/vscode-appland/blob/develop/package.json#L473

Then any updates to this file will be pushed via semantic-release
https://github.com/getappmap/vscode-appland/blob/develop/release.config.js#L27

dustinbyrne (approved):
I haven't run this locally yet but the change set looks good

resources/versions.json
kgilpin: OK, I had these changes locally so I thought I should push them. I will leave it alone if it doesn't do any harm.

PR #949, opened 2024-04-22
feat: Integrate the context pane

dustinbyrne:
image

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -387,3 +387,19 @@ export function sanitizeEnvironment(env: NodeJS.ProcessEnv): Record<string, stri
   }
   return sanitizedEnv;
 }
+
+export function parseLocation(location: string): vscode.Location | vscode.Uri {
+  const match = location.match(/:(\d+(-\d+)?)$/);
+  if (match) {
+    const path = location.substring(0, match.index);
+    const [startLine, endLine] = match[1].split('-').map((x) => parseInt(x, 10) - 1);
+    return new vscode.Location(
+      vscode.Uri.parse(path),
+      new vscode.Range(
+        new vscode.Position(startLine, 0),
+        new vscode.Position(endLine ?? startLine, 0)
+      )
+    );
+  }
+  return vscode.Uri.parse(location);
+}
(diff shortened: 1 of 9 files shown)


src/webviews/chatSearchWebview.ts
```diff
@@ -137,6 +137,35 @@ export default class ChatSearchWebview {
           await vscode.commands.executeCommand('vscode.openWith', uri, 'appmap.views.appMapFile');
           break;
         }
+        case 'open-location': {
+          const { location } = message;
+          const [path, lineNumbers] = location.split(':');
```
ahtrotta, line 142: If this is an absolute path in Windows with a drive letter then this won't work correctly

dustinbyrne: Good catch! This is now fixed.

test/unit/util.test.ts
```diff
@@ -32,3 +34,36 @@ describe('sanitizeEnvironment', () => {
...
+
+      expect(result instanceof URI).to.be.true;
+      expect(result.fsPath).to.equal(expected);
+    });
+  });
+
+  it('should parse a location with a line number', () => {
+    const examples = [
+      ['/path/to/file.rb:11', '/path/to/file.rb', 10],
+      ['C:\\path\\to\\file.rb:11', '\\path\\to\\file.rb', 10],
```
ahtrotta, line 56: Is it okay that the drive letter gets removed? If someone has multiple drives would VS Code be confused?

src/util.ts
```diff
@@ -387,3 +387,19 @@ export function sanitizeEnvironment(env: NodeJS.ProcessEnv): Record<string, stri
   }
   return sanitizedEnv;
 }
+
+export function parseLocation(location: string): vscode.Location | vscode.Uri {
+  const match = location.match(/:(\d+(-\d+)?)$/);
+  if (match) {
+    const path = location.substring(0, match.index);
+    const [startLine, endLine] = match[1].split('-').map((x) => parseInt(x, 10) - 1);
```
ahtrotta, line 395: The number after the colon for a data request is an event ID for an AppMap, so we need to handle that differently, right?

ahtrotta (approved):
Looks good and worked well when I tested it manually! :+1:

Issue #950, opened 2024-04-24
AppMap: Open as JSON fails when run from the command palette


apotterri:
When viewing an AppMap, running `AppMap: Open as JSON` from the command palette fails with an exception:
<img width="1842" alt="image" src="https://github.com/getappmap/vscode-appland/assets/4212483/e85c684f-c612-4a56-82f3-5da10ffbb12f">

PR #951, opened 2024-04-24
Fix openAsJson

apotterri:
Make sure openAsJson doesn't fail when run from the command palette.

Fixes: #950


src/tree/contextMenu.ts
```diff
@@ -26,7 +26,20 @@ export default class ContextMenu {
...
-        vscode.commands.executeCommand('vscode.openWith', item.descriptor.resourceUri, 'default');
+        let uri: vscode.Uri;
+        if (!item) {
+          const { activeTab } = vscode.window.tabGroups.activeTabGroup;
+          if (!activeTab) {
+            vscode.window.showErrorMessage('No active editor.');
+            return;
+          }
+
+          uri = (activeTab.input as vscode.TabInputCustom).uri;
```
dividedmind, line 37: Should check the type:
```typescript
const input = vscode.window.tabGroups.activeTabGroup.activeTab?.input;
if (input instanceof vscode.TabInputCustom && input.viewType === ''appmap.views.appMapFile')
  uri = input.uri;
else {
  vscode.window.showErrorMessage('No AppMap open in the active tab.');
  return;
}
```

dividedmind (changes requested):

test/integrationTest.ts
```diff
@@ -41,11 +41,18 @@ async function integrationTest() {
   const projectRootDir = resolve(__dirname, '..');
   const testDir = resolve(__dirname, '../out/test/integration');
 
-  let fileArgs = process.argv.slice(1).filter((arg) => arg.match(/\.test\.(?:js|ts)$/));
-  if (fileArgs.length > 0) {
+  let fileArgs = process.argv.slice(1);
+  if (fileArgs.length > 1) {
+    const matchedArgs = fileArgs.filter((arg) => arg.match(/\.test\.(?:js|ts)$/));
+    if (matchedArgs.length === 0) {
+      throw new Error(`No test files matched ${fileArgs}`);
```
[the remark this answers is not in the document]
apotterri: No, it doesn't. This does, though: https://github.com/getappmap/vscode-appland/blob/0b62e1e9d3eb4b3452b95a9dbaeb8bfb99ffc46e/test/integrationTest.ts#L68 .

src/tree/contextMenu.ts
apotterri: I cribbed this code from the `deleteAppMap` command, which also didn't check the type, and went ahead and deleted the file in the active tab without confirmation(!).

I added the check there, too.

src/tree/contextMenu.ts
```diff
@@ -34,6 +34,17 @@ export default class ContextMenu {
...
+            input instanceof vscode.TabInputCustom &&
+            input.viewType === 'appmap.views.appMapFile'
+          )
+            uri = input.uri;
+          else {
+            vscode.window.showErrorMessage('No AppMap open in the active tab.');
+            return;
+          }
+
           uri = (activeTab.input as vscode.TabInputCustom).uri;
```
dividedmind, line 48: This line can probably go now.

src/tree/contextMenu.ts
```diff
@@ -86,6 +97,17 @@ export default class ContextMenu {
...
```
dividedmind, line 111: Ditto.

dividedmind (approved):
Perhaps it'd be a good idea to extract the common code into a function?

src/tree/contextMenu.ts
apotterri: Oops, fixed. Thanks.

dividedmind:
BTW, another thing I noticed — the signature should probably now make `item` optional.

PR #953, opened 2024-04-26
fix: Retry downloads when needed, show progress and allow cancellation

### LINKED ISSUE #944: Handle partial downloads
Partial downloads are problematic in that they:
- will not be detected
- won't be retried
- will have a symlink pointed at them upon restarting the editor

There are a few ways to solve this:
- downloads can be verified via sha sum upon initialization
- partial downloads are not written to their final destination on disk until the download has completed
  - this should force downloads to restart if the window was closed before a download completed

dividedmind:
Fixes #944

— Downloads files first to `filename.part` target; only renames them after the download is successful.
— On failure, the download is retried (up to 3 times, with 5, 25, 125 second backoff).
— Assets are downloaded sequentially, not simultaneously.
— Progress notification is shown in the UI:
download
— The user can cancel the operation. This will cancel downloading of assets completely. If there was a previous version installed, it'll keep being used.
— Show an error message to the user on error.
— Don't replace a symlink if we're not updating and it's present and not broken. This will allow the user to override the binary by creating their own symlink and not get it clobbered by the asset service.
— Redownload all the assets unconditionally unless `.appmap/.download-complete` exists to fix any old broken state. Create this file on successful install of assets.


kgilpin:
Can I get a bit more in the Description?  This is a big diff. I’d like some high level idea of what’s going on. Thx.

dividedmind:
One improvement that comes to mind: don't replace a symlink if we're not updating and it's present and not broken. This will allow the user to override the binary by creating their own symlink and not get it clobbered by the asset service.

kgilpin:
Since we will trust this to repair a broken download, I suggest that we re-install the assets the first time a user upgrades to this version of the extension.

dividedmind:
> Since we will trust this to repair a broken download, I suggest that we re-install the assets the first time a user upgrades to this version of the extension.

Done.

dustinbyrne (approved):

PR #954, opened 2024-04-29
fix: Include `directory` when parsing context location

dustinbyrne:


ahtrotta:
I re-ran the integration tests and a different one failed, so it seems like they're suddenly unstable again. Both failing tests had to do with expecting tree providers to have the expected tree items.

kgilpin:
> I re-ran the integration tests and a different one failed, so it seems like they're suddenly unstable again. Both failing tests had to do with expecting tree providers to have the expected tree items.

FYI tree provider has a known bug - https://github.com/getappmap/vscode-appland/pull/930

ahtrotta (approved):
Looks good and worked well when I tested it manually in a workspace with multiple project folders :+1:

PR #974, opened 2024-06-05
feat: Scanner is disabled by default

### LINKED ISSUE #973: Scanner is disabled by default
The scanner process should no longer be run by default. Instead, a toggle in settings should be added to enable it.

zermelo-wisen:
Fixes #973.

- Adds "Enable AppMap scanner" to the extension settings to enable/disable `scanner` process with the default value of `disabled`.
- Change is applied only after restart.
- Adds a test to verify the new setting.
- Adds `scannerEnabled: true` setting to the fixture projects which are used in the tests assuming `scanner` is running by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -297,6 +297,11 @@
         "appMap.navie.rpcPort": {
           "type": "number",
           "description": "Port number to pass to the Navie UI (used for extension development and debugging only)"
+        },
+        "appMap.scannerEnabled": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable AppMap scanner"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "qna": "https://appmap.io/docs/faq.html",
   "engines": {
-    "vscode": "^1.82.0"
+    "vscode": "^1.61.0"
   },
   "agents": {
     "ruby": ">=0.60.0"
--- a/package.json
+++ b/package.json
@@ -394,7 +394,7 @@
           "id": "appmap.views.findings",
           "name": "Runtime Analysis",
           "visibility": "collapsed",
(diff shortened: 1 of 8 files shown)


dustinbyrne:
@zermelo-wisen, this change looks great. There's a 'Runtime Analysis' tree view panel which renders the results of the scanner process in the AppMap sidebar. Could you add this to the toggle as well?
image

Perhaps we can just change the visibility to `hidden`? 
https://code.visualstudio.com/api/extension-guides/tree-view#contributing-views-to-view-containers

The actual tree is declared here, if it's helpful.
https://github.com/getappmap/vscode-appland/blob/develop/src/tree/findingsTreeDataProvider.ts

zermelo-wisen:
> @zermelo-wisen, this change looks great. There's a 'Runtime Analysis' tree view panel which renders the results of the scanner process in the AppMap sidebar. Could you add this to the toggle as well? image
> 
> Perhaps we can just change the visibility to `hidden`? https://code.visualstudio.com/api/extension-guides/tree-view#contributing-views-to-view-containers
> 
> The actual tree is declared here, if it's helpful. https://github.com/getappmap/vscode-appland/blob/develop/src/tree/findingsTreeDataProvider.ts

It's not possible to change visibility property in code. It is specified in `package.json`. If I skip creating view in initialization, the resulting view looks like in the image below.

```TypeScript
  const findingsTreeProvider = new FindingsTreeDataProvider(context, appmapCollection);
  if (ExtensionSettings.scannerEnabled) {
    vscode.window.createTreeView('appmap.views.findings', {
      treeDataProvider: findingsTreeProvider,
    });
  }
  context.subscriptions.push(findingsTreeProvider);

```

image

Should we do something like EmptyDataProvider or provide only one root item with the text like "You have to enable scanner to see the results here."?

PR #976, opened 2024-06-12
fix: The authentication provider is initialized before it is needed

dustinbyrne:
Otherwise, depending on other initialization tasks, it may be possible for a timeout to occur while waiting for the provider to register.


dustinbyrne:
I'm not confident this fixes the issue.

PR #977, opened 2024-06-12
fix: Authentication provider timeouts

dustinbyrne:
This change handles cases where `vscode.authentication.getSession` may time out waiting for the authentication provider to register. Upon initialization, we immediately request a session, however we don't initialize the authentication provider until we're halfway through activation.

Just handling this error is enough to fix the issue, however I've also moved the authentication provider's initialization to the top of activation to fix a cosmetic issue. On its own, this may have been enough to fix the issue, but catching the exception is a guarantee.

### DIFF
--- a/src/authentication/index.ts
+++ b/src/authentication/index.ts
@@ -10,8 +10,16 @@ export const AUTHN_PROVIDER_NAME = 'appmap.server';
 export async function getApiKey(createIfNone: boolean): Promise<string | undefined> {
   if (!createIfNone && Environment.appMapTestApiKey) return Environment.appMapTestApiKey;
 
-  const session = await vscode.authentication.getSession(AUTHN_PROVIDER_NAME, ['default'], {
-    createIfNone,
-  });
+  let session: vscode.AuthenticationSession | undefined;
+  try {
+    session = await vscode.authentication.getSession(AUTHN_PROVIDER_NAME, ['default'], {
+      createIfNone,
+    });
+  } catch (e) {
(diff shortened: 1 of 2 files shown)


dustinbyrne:
I was able to verify this works by temporarily adding the registration of our authentication provider into a `setTimeout` block, exceeding VS Codes timeout period of 5s.

dustinbyrne:
The source of the exception: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/authentication/browser/authenticationService.ts#L220-L226

src/authentication/index.ts
```diff
@@ -10,8 +10,16 @@ export const AUTHN_PROVIDER_NAME = 'appmap.server';
...
+  let session: vscode.AuthenticationSession | undefined;
+  try {
+    session = await vscode.authentication.getSession(AUTHN_PROVIDER_NAME, ['default'], {
+      createIfNone,
+    });
+  } catch (e) {
+    // VSCode may throw a string instead of an Error, e.g., if the authentication provider is not registered in time.
+    const err = e instanceof Error ? e : new Error(String(e));
+    console.error(err);
+  }
```
dustinbyrne, line 22: This doesn't incur any side effects from what I've observed. Anything that needs an API key is typically listening for it to change anyway.

apotterri (approved):
Getting this out seems pretty important, so let's go with this.

If there's some test we can add to make sure we don't run afoul of the timeout in the future, we might that, too.

PR #1009, opened 2024-08-28
feat: Implement cleanup of old AppMap binaries

### LINKED ISSUE #1007: Delete old AppMap binaries
### Problem
The AppMap plugin currently does not clean up older versions of AppMap binaries located in the `$HOME/.appmap` (Linux/macOS) or `%HOME%/.appmap` (Windows) directory. This can lead to unnecessary storage usage as multiple old versions accumulate over time. The goal is to remove all older versions of the binaries and ensure that only the most recent version is retained moving forward.

### Analysis
To resolve this issue, we need to implement a cleanup function that scans the AppMap binaries directory and deletes all but the most recent binary. The most recent binary can be determined by the semver version number in the file name. 

Furthermore, we should integrate this cleanup function to be automatically executed whenever a new AppMap binary is downloaded or installed. This ensures that the directory is consistently maintained with only the latest version, preventing storage bloat.

zermelo-wisen:
Fixes #1007.

- Added 'cleanUpOldBinaries' and 'cleanUpOldBinariesFromDir' functions to handle the removal of older binaries.
- Integrated 'cleanUpOldBinaries' to run after downloading a binary.
- Added a unit test to verify the cleanup process.

This change addresses storage bloat by ensuring that only the latest versions of binaries are kept, enhancing the efficiency and storage management of the AppMap plugin.

### Problem

The AppMap plugin currently does not clean up older versions of AppMap binaries located in the `$HOME/.appmap` (Linux/macOS) or `%HOME%/.appmap` (Windows) directory. This can lead to unnecessary storage usage as multiple old versions accumulate over time. The goal is to remove all older versions of the binaries and ensure that only the most recent version is retained moving forward.

### Analysis

To resolve this issue, we need to implement a cleanup function that scans the AppMap binaries directory and deletes all but the most recent binary. The most recent binary can be determined by the `semver` version number in the file name.

### Solution

This PR introduces:

1. **Cleanup Function**: A function that removes all but the latest version of the binaries in the `$HOME/.appmap` directory.
2. **Automation**: The cleanup function is automatically executed whenever a new AppMap binary is downloaded or during start up.

### Code Changes

**1. Directory Management and Utility Enhancements**
  - Introduced `getAppmapDir` to centralize the base directory path for `.appmap`.

**2. Cleanup Logic Implementation**
  - Added `cleanUpOldBinaries` and `cleanUpOldBinariesFromDir` to handle the removal of older binaries.

**3. Integration**
  - Integrated `cleanUpOldBinaries` to run after downloading a binary.

**Modifications**:

```typescript
import * as fs from 'fs';
import * as path from 'path';
import * as semver from 'semver';

...
export default class AssetService {
   ...
  public static async updateAll(throwOnError = false): Promise<void> {
    ...
    await this.cleanUpOldBinaries();
    ...
  }

  public static async updateOne(assetId: AssetIdentifier): Promise<void> {
    ...
    await this.cleanUpOldBinaries();
    return result;
  }

  static async cleanUpOldBinaries() {
    await this.cleanUpOldBinariesFromDir(join(AssetService.getAppmapDir(), 'lib', 'appmap'));
    await this.cleanUpOldBinariesFromDir(join(AssetService.getAppmapDir(), 'lib', 'java'));
    await this.cleanUpOldBinariesFromDir(join(AssetService.getAppmapDir(), 'lib', 'scanner'));
  }

  private static async cleanUpOldBinariesFromDir(directory: string) {
    try {
      const files = await fs.promises.readdir(directory);
      const fileGroups: { [key: string]: { name: string; version: string }[] } = {};

      files.forEach((file) => {
        const version = this.extractVersion(file);
        if (version && semver.valid(version)) {
          const groupName = file.replace(version, '');
          if (!fileGroups[groupName]) {
            fileGroups[groupName] = [];
          }
          fileGroups[groupName].push({ name: file, version });
        }
      });

      for (const groupName in fileGroups) {
        // Sort in descending order of versions
        const binaries = fileGroups[groupName].sort((a, b) => semver.compare(b.version, a.version));

        // Keep the latest version, delete the rest
        for (let i = 1; i < binaries.length; i++) {
          const filePath = path.join(directory, binaries[i].name);
          await fs.promises.unlink(filePath);
          log.info(`Deleted old binary: ${binaries[i].name}`);
        }
      }
    } catch (error) {
      log.error(`Failed to clean up old binaries: ${error}`);
    }
  }

  private static extractVersion(fileName: string): string | null {
    // Match patterns like '-v1.2.3', '-v1.2.3-alpha', '-v1.2.3+build', '-1.2.3-alpha+build'
    // with or without file extension
    const versionMatch = fileName.match(
      /-v?(\d+\.\d+\.\d+(-[a-zA-Z0-9-]+)?(\+[a-zA-Z0-9-]+)?)\b(\.[^.]+)?$/
    );
    return versionMatch ? versionMatch[1] : null;
  }
}
```

**4. Testing**
- Added a unit test to verify the cleanup process.

**Test Modifications**:

```typescript
it('cleans up old binaries', async () => {
  ...
  await AssetService.updateAll(true);

  // Verify the results
  const verifyFiles = async (directory: string, expectedFiles: string[]) => {
    const files = await readdir(join(homeDir, '.appmap', directory));
    const expectedSet = new Set(expectedFiles);
    const remainingSet = new Set(files);

    expect(expectedSet).to.deep.equal(remainingSet);
  };

  // Verify each directory's expected files
  await verifyFiles(join('lib', 'appmap'), ['appmap-v0.0.0-TEST']);
  await verifyFiles(join('lib', 'java'), [
    'appmap-1.26.2.jar',
    'appmap-agent-1.26.2.jar',
    'appmap.jar',
  ]);
  await verifyFiles(join('lib', 'scanner'), ['scanner-v1.88.0']);
});
```

### PR Summary

This PR addresses storage bloat by implementing cleanup functionality for old binaries in the AppMap plugin. This ensures that only the latest versions of binaries are kept, enhancing the efficiency and storage management of the AppMap plugin. This update is crucial to maintain optimal performance and prevent unnecessary storage consumption.

### DIFF
--- a/src/assets/assetService.ts
+++ b/src/assets/assetService.ts
@@ -1,3 +1,7 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as semver from 'semver';
+
 import { AbortError } from 'node-fetch';
 import * as vscode from 'vscode';
 import {
@@ -21,6 +25,10 @@ export enum AssetIdentifier {
   ScannerCli,
   JavaAgent,
 }
+
+// Order is important while checking with fileName.startsWith
+const binaryPrefixes = ['appmap-agent-', 'appmap-', 'scanner-'];
+const binarySuffixes = ['.exe', '.jar'];
 
 export default class AssetService {
   private static _extensionDirectory: string;
@@ -61,22 +69,26 @@ export default class AssetService {
     }
   }
 
+  public static getAppmapDir() {
+    return join(homedir(), '.appmap');
+  }
+
   public static getAssetPath(assetId: AssetIdentifier): string {
     // This could be a property on each AssetDownloader
     switch (assetId) {
       case AssetIdentifier.AppMapCli:
-        return join(homedir(), '.appmap', 'bin', binaryName('appmap'));
+        return join(this.getAppmapDir(), 'bin', binaryName('appmap'));
       case AssetIdentifier.ScannerCli:
-        return join(homedir(), '.appmap', 'bin', binaryName('scanner'));
+        return join(this.getAppmapDir(), 'bin', binaryName('scanner'));
       case AssetIdentifier.JavaAgent:
-        return join(homedir(), '.appmap', 'lib', 'java', 'appmap.jar');
+        return join(this.getAppmapDir(), 'lib', 'java', 'appmap.jar');
       default:
         throw new Error(`Invalid asset ID ${assetId}`);
     }
   }
 
   public static async updateAll(throwOnError = false): Promise<void> {
-    const appmapDir = join(homedir(), '.appmap');
+    const appmapDir = this.getAppmapDir();
     const dirs = [join(appmapDir, 'bin'), join(appmapDir, 'lib')];
     await Promise.all(dirs.map((dir) => mkdir(dir, { recursive: true })));
 
@@ -116,13 +128,14 @@ export default class AssetService {
               sync.emit('error', e);
               if (e instanceof AbortError) return reject(e);
             }
+          await this.cleanUpOldBinaries();
           await initialDownloadCompleted();
         });
     });
   }
 
   public static async updateOne(assetId: AssetIdentifier): Promise<void> {
-    const appmapDir = join(homedir(), '.appmap');
+    const appmapDir = this.getAppmapDir();
     const sync = new LockfileSynchronizer(appmapDir);
     return new Promise<void>((resolve, reject) => {
       sync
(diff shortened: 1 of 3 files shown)


src/assets/assetService.ts
```diff
@@ -113,8 +122,58 @@ export default class AssetService {
...
+
+      for (const groupName in fileGroups) {
+        // Sort in descending order of versions
+        const binaries = fileGroups[groupName].sort((a, b) => semver.compare(b.version, a.version));
+
+        // Keep the latest version, delete the rest
+        for (let i = 1; i < binaries.length; i++) {
+          const filePath = path.join(directory, binaries[i].name);
+          await fs.promises.unlink(filePath);
+          log.info(`Deleted old binary: ${binaries[i].name}`);
```
dustinbyrne, line 163: I think it's worth some sort of assertion that we're not destroying a linked object. These binaries are put in the lib directory, then a symlink is created in the bin directory pointing at the latest binary in lib.

It seems unlikely that this would happen, but if for some reason an asset download failed and the symlink wasn't updated, it may be possible that the linked binary gets deleted here.

dustinbyrne (changes requested):
This is looking good. Please see the requested changes above. Also, the build is failing but it's not clear to me that it's related. Could you look in to it?

zermelo-wisen:
> This is looking good. Please see the requested changes above. Also, the build is failing but it's not clear to me that it's related. Could you look in to it?

Changes are done. The integration test is failing, but I re-ran the build of the last release commit, and it seems to be failing as well. This is probably unrelated, but I can try to solve it even if it's unrelated.

src/assets/assetService.ts
```diff
@@ -113,8 +122,80 @@ export default class AssetService {
...
+    } catch (error) {
+      log.error(`Failed to clean up old binaries: ${error}`);
+    }
+  }
+
+  private static extractVersion(fileName: string): string | null {
+    // Match patterns like '-v1.2.3', '-v1.2.3-alpha', '-v1.2.3+build', '-1.2.3-alpha+build'
+    // with or without file extension
+    const versionMatch = fileName.match(
+      /-v?(\d+\.\d+\.\d+(-[a-zA-Z0-9-]+)?(\+[a-zA-Z0-9-]+)?)\b(\.[^.]+)?$/
```
kgilpin, line 197: Ensure that we are using a recommended RegExp from here - https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

src/assets/assetService.ts
```diff
@@ -94,13 +102,14 @@ export default class AssetService {
               sync.emit('error', e);
               if (e instanceof AbortError) return reject(e);
             }
+          await this.cleanUpOldBinaries();
```
kgilpin, line 105: Trap, log, and ignore any errors that occur here, so that the cleanup doesn't interfere with any other update functionality.

src/assets/assetService.ts
```diff
@@ -113,8 +122,80 @@ export default class AssetService {
...
+        for (let i = 1; i < binaries.length; i++) {
+          const binaryPath = path.join(dir, binaries[i].name);
+
+          // Check if the binary is the target of any symlink
+          if (hasSymlink(binaryPath)) {
+            log.warning(`Skipping deletion of ${binaries[i].name} as it's targeted by a symlink.`);
+            continue;
+          }
+
+          await fs.promises.unlink(binaryPath);
```
kgilpin, line 184: Other processes may be concurrently cleaning up the same binaries. Therefore, there is a race condition here.

src/assets/assetService.ts
zermelo-wisen: I was doing it inside the inner most function `AssetService.cleanUpOldBinariesFromDir`, which is called for `appmap`, `java` and `scanner` folders separately. In order to cover symlink resolutions as well, now I moved the try catch logic to `AssetService.cleanUpOldBinaries` instead of the call site above, because `AssetService.cleanUpOldBinaries` is called from two different places.

dustinbyrne:
Closing this for now. The plan is to instead implement this within the appmap CLI.

PR #1015, opened 2024-09-10
Add the ability to pin files to the Navie Context

apotterri:
Fixes https://github.com/getappmap/appmap-js/issues/1951 .


src/webviews/chatSearchWebview.ts
```diff
@@ -56,6 +58,49 @@ export default class ChatSearchWebview {
     return this.webviewList.currentWebview;
   }
 
+  async doPinFiles(panel: vscode.WebviewPanel, reqs: PinFileRequest[]) {
+    const maxPinnedFileSize = ExtensionSettings.maxPinnedFileSize;
+    assert(maxPinnedFileSize, 'maxPinnedFileSize unset?');
```
dustinbyrne, line 63: I don't believe it can be unset given the default, but this would raise on `0`.

src/webviews/chatSearchWebview.ts
```diff
@@ -190,6 +235,23 @@ export default class ChatSearchWebview {
...
+          await this.chooseFilesToPin().then(async (uris: vscode.Uri[] | undefined) => {
+            if (!uris) return;
+            const requests: PinFileRequest[] = uris.map((u) => ({
+              name: u.toString(),
+              uri: u.toString(),
+            }));
+            this.doPinFiles(panel, requests);
+          });
+          break;
+        }
```
dustinbyrne, line 248: I don't see this message emitted from the frontend

src/webviews/chatSearchWebview.ts
apotterri: Fixed.

src/webviews/chatSearchWebview.ts
apotterri: You're right, it's not. It gets posted here: https://github.com/getappmap/vscode-appland/pull/1015/files#diff-f0135d8b0aeee6d012852ba50800f465cb5adb311dfac30cd79eb6ed99063882R31 .

I implemented the command this way because we need to have VS Code open the File Dialog for the user to choose the files to pin and get the content of the files. Then, we need to tell the Vue component to actually pin the content to the Context. I couldn't just add a method to `chatSearchWebview`, because `panel` is local to `explain`: https://github.com/getappmap/vscode-appland/blob/pin-files_20240905/src/webviews/chatSearchWebview.ts#L123

dustinbyrne (approved):

PR #1028, opened 2024-10-01
fix: Update the VS Code marketplace extension readme. 

petecheslock:
@kgilpin called out missing commands which I added.  

Do we want to update the examples?  We've got an end to end Navie video now - do we want to embed or link to this video? 
https://vimeo.com/1012769182

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ AppMap is a developer tools platform powered by runtime analysis and AI. It's co
 * AppMap data - Runtime traces of your application runtime.
 * AppMap diagrams - Interactive visualizations of AppMap data.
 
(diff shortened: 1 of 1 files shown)


Daniel-Warner-X (commented):
@petecheslock I think you can remove the numbered list of links to specific examples and update the text in the 'Examples' section to make 'Here are some examples...' a link that goes to https://appmap.io/product/examples/navie.

Also, I think you can remove 'By using AppMap data...' on line 14 and start that paragraph at `Navie is the first...`. That initial clause makes the sentence hard to parse.

Daniel-Warner-X (approved):

PR #1031, opened 2024-10-03
fix: Write an OpenAI-compatible error when context overflows

dustinbyrne:
Previously the server would report an error stating `Message exceeds token limit.`.

This error is now transformed into an OpenAI API compatible error, mimicking the same error returned by the official OpenAI API. This change allows Navie to properly reduce the token count before trying again.

### DIFF
--- a/test/unit/services/chatCompletion.test.ts
+++ b/test/unit/services/chatCompletion.test.ts
@@ -18,8 +18,8 @@ const mockModel: LanguageModelChat = {
   id: 'test-model',
   family: 'test-family',
   version: 'test-model',
-  async countTokens(): Promise<number> {
-    return 0;
+  async countTokens(content: string): Promise<number> {
+    return content.length;
   },
   maxInputTokens: 325,
   name: 'Test Model',
@@ -163,12 +163,13 @@ describe('ChatCompletion', () => {
   });
 
   describe('when streaming errors', () => {
+    let error = new Error('test error');
     beforeEach(() => {
       mockModel.sendRequest = async () => {
         return {
           // eslint-disable-next-line require-yield
           text: (async function* () {
-            throw new Error('test error');
+            throw error;
           })(),
         };
       };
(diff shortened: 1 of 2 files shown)


test/unit/services/chatCompletion.test.ts
```diff
@@ -189,13 +190,45 @@ describe('ChatCompletion', () => {
...
     it('reports errors when not streaming', async () => {
+      const messages = [
+        { content: 'Hello', role: 'user' },
+        { content: 'How are you?', role: 'assistant' },
+        { content: 'I am good, thank you!', role: 'user' },
+      ];
+      const response = await postAuthorized(chatCompletion.url, {
+        model: 'test-model',
+        messages,
+        stream: true,
```
apotterri, line 201: `stream` shouldn't be here, should it? The example says, "when not streaming".

test/unit/services/chatCompletion.test.ts
```diff
@@ -189,13 +190,45 @@ describe('ChatCompletion', () => {
...
+    });
+
+    it('reports token usage when exceeded (streaming)', async () => {
+      error = new Error('Message exceeds token limit.');
+      const messages = [
+        { content: 'Hello', role: 'user' },
+        { content: 'How are you?', role: 'assistant' },
+        { content: 'I am good, thank you!', role: 'user' },
+      ];
+      const response = await postAuthorized(chatCompletion.url, { model: 'test-model', messages });
```
apotterri, line 213: And this one seems like it needs `stream: true`?

apotterri (changes requested):
There are a couple of cases where the value of `stream` doesn't seem like it matches the example description.

test/unit/services/chatCompletion.test.ts
dustinbyrne: Yeah, it looks like I had named the test cases backwards. This is fixed now :+1:

apotterri (approved):

dividedmind (approved):

PR #1039, opened 2024-10-15
Feat/copilot UI integration

dustinbyrne:
This PR:
- Integrates the updated Copilot UI (https://github.com/getappmap/appmap-js/pull/2059)
- Reloads Navie without requiring a new window or reload of the extension
- Binds the Copilot proxy to loopback via the IPv4 address


src/services/chatCompletion.ts
```diff
@@ -75,7 +79,7 @@ export default class ChatCompletion implements Disposable {
   }
 
   get url(): string {
-    return `http://localhost:${this.port}/vscode/copilot`;
+    return `http://${this.host}:${this.port}/vscode/copilot`;
```
dividedmind, line 82: FWIW, this isn't entirely correct — if you pass '0.0.0.0' or '::' as `host` it will bind correctly but the URL will be invalid. Not that it matters much here ;)

PR #1045, opened 2024-10-25
Update walkthrough

Daniel-Warner-X:
This PR is part of the [Get New Users into Chat](https://github.com/getappmap/board/issues/335) epic and updates two views in the walkthrough.

The first is the screen a user sees when they are logged out. Initially I thought I could remove the image but vscode requires the walkthrough to have an image, so I put in a screenshot prompting the user to open the sidebar and expose the sign-in form. I also updated the language on this view to be consistent with  the concepts of 'activate' and 'sign-in.'

Also, I thought I could change the headline on this page to be 'Activate AppMap' but the headline remains constant throughout the entire walkthrough so I kept it as is 'Get Started with AppMap Navie.'

Updated view:

Screenshot 2024-10-24 at 7 49 19 PM

Current view:

image

The second screen that I updated is the 'Choose your LLM' screen-- the second step in the walkthrough. This is the screen that users are much more likely to see. Here I updated the text to reflect the fact that we currently default to GH CoPilot if it is installed. I also added an image to prompt the user to change their LLM from the Navie Chat if they need to.

Updated view:
Screenshot 2024-10-24 at 7 48 47 PM

Current view:
image


dividedmind (approved):
Looks good. I removed an unused image; please remember to squash merge.

PR #1053, opened 2024-11-21
Select Copilot model and default to `gpt-4o`

dustinbyrne:
This PR adds a new command `AppMap: Select Copilot Model` that allows users to change their preferred Copilot model. The default model is set to `gpt-4o`.

## Changes
- Added a new command `appmap.copilot.selectModel` that lets users pick from available Copilot models
- Added configuration setting `appMap.copilot.preferredModel` to store the user's model preference
- Set default model to `gpt-4o`
- The RPC server restarts after model selection to apply changes
- Added unit tests to verify the model selection functionality

## Testing
The changes have been tested to verify:
- Model selection UI works correctly
- Setting is saved properly
- RPC server restarts after model change
- Error handling for when no models are available
- Cancellation handling for the model picker

## Notes
The command can be accessed via:
- Command palette: `AppMap: Select Copilot Model`
- Default model is `gpt-4o` if no preference is set

Opening for review - please let me know if any questions or changes needed.

### DIFF
--- a/src/commands/pickCopilotModel.ts
+++ b/src/commands/pickCopilotModel.ts
@@ -1,0 +1,32 @@
+import ExtensionSettings from '../configuration/extensionSettings';
+import ChatCompletion from '../services/chatCompletion';
+import vscode from 'vscode';
+
+export default class PickCopilotModelCommand {
+  public static readonly command = 'appmap.copilot.selectModel';
+
+  public static register(context: vscode.ExtensionContext): void {
+    context.subscriptions.push(
+      vscode.commands.registerCommand(
+        PickCopilotModelCommand.command,
+        PickCopilotModelCommand.execute
+      )
+    );
+  }
+
+  public static async execute(): Promise<void> {
+    await ChatCompletion.refreshModels();
+    if (ChatCompletion.models.length === 0) {
+      vscode.window.showErrorMessage('No Copilot models are available.');
+      return;
+    }
+
+    const model = await vscode.window.showQuickPick(
+      ChatCompletion.models.map((m) => ({ label: m.name, details: m.id }))
+    );
+    if (!model) return;
+
+    await ExtensionSettings.setPreferredCopilotModel(model.details);
+    return vscode.commands.executeCommand('appmap.rpc.restart');
+  }
+}
--- a/src/services/chatCompletion.ts
+++ b/src/services/chatCompletion.ts
@@ -106,7 +106,10 @@ export default class ChatCompletion implements Disposable {
     return env;
   }
 
-  private static models: vscode.LanguageModelChat[] = [];
+  private static _models: vscode.LanguageModelChat[] = [];
+  public static get models(): ReadonlyArray<vscode.LanguageModelChat> {
+    return this._models;
+  }
 
   static get preferredModel(): vscode.LanguageModelChat | undefined {
     return ChatCompletion.models[0];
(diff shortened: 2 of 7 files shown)


kgilpin:
Hi @dustinbyrne thanks! I have taken a look, I think @dividedmind should take a look too, just to be sure we don't miss anything.

src/commands/pickCopilotModel.ts
```diff
@@ -0,0 +1,32 @@
...
+      return;
+    }
+
+    const model = await vscode.window.showQuickPick(
+      ChatCompletion.models.map((m) => ({ label: m.name, details: m.id }))
+    );
+    if (!model) return;
+
+    await ExtensionSettings.setPreferredCopilotModel(model.details);
+    return vscode.commands.executeCommand('appmap.rpc.restart');
```
dividedmind, line 30: I would add watching for this configuration change at https://github.com/getappmap/vscode-appland/pull/1053/files#diff-3606d80ad0d196e57c8e0c0bab40cdfa917775d2b2f4074a5a8ecdb04f778eb2R239 instead. This way there's only one code path there that takes care of all the corner cases (including restarting the RPC only when copilot mode is enabled and the model actually was changed).

src/services/chatCompletion.ts
```diff
@@ -106,15 +106,23 @@ export default class ChatCompletion implements Disposable {
...
+  private static _models: vscode.LanguageModelChat[] = [];
+  public static get models(): ReadonlyArray<vscode.LanguageModelChat> {
+    return this._models;
+  }
 
   static get preferredModel(): vscode.LanguageModelChat | undefined {
-    return ChatCompletion.models[0];
+    const modelId = ExtensionSettings.preferredCopilotModel;
+    if (modelId) {
+      const model = this.models.find((m) => m.id === modelId);
```
dividedmind, line 117: This is backwards — `this._models` is refreshed in `refreshModels()` after `preferredModel()` is called, so `preferredModel()` can use stale data. Maybe `refreshModels()` could take the preference into account when sorting instead?

test/unit/commands/pickCopilotModel.test.ts
```diff
@@ -0,0 +1,67 @@
...
+          maxInputTokens: 325,
+          family: 'copilot',
+        },
+      ];
+      showQuickPickStub = sinon.stub(vscode.window, 'showQuickPick').callsFake(() => ({
+        details: chosenModel,
+      }));
+      showErrorMessageStub = sinon.stub(vscode.window, 'showErrorMessage').resolves();
+      executeCommandStub = sinon.stub(vscode.commands, 'executeCommand').resolves();
+      sinon.stub(vscode.lm, 'selectChatModels').callsFake(() => models as any); // eslint-disable-line @typescript-eslint/no-explicit-any
```
dividedmind, line 28: There is mock support in `test/unit/mock/vscode/lm.ts` — did you miss it or is there a reason not to use it here?

dividedmind (changes requested):
Looks mostly good, but that reversed data dependency in `preferredModel()` and `refreshModels()` can lead to weird bugs and should be fixed.

src/services/chatCompletion.ts
dividedmind: BTW, is there a reason to change the model list to per-instance? It is a global state after all.

dustinbyrne: Thanks, I'll respond to each point in order:
1. Wasn't `preferredModel` always using stale data then? My understanding is that `refreshModels` is calling this method before the refresh to determine whether or not it changed. That seems logically sound. [When the preferred model is consumed](https://github.com/getappmap/vscode-appland/blob/f7423b43657dba8dbe64b13f83d93d7cc4babae4/src/services/chatCompletion.ts#L279-L280), it's retrieved following a refresh. Maybe I'm missing your point.
2. This is still a static method, I've not changed anything about the data model.

dustinbyrne: Regarding #1, I believe that I understand what you're saying now. If I'm restarting the service by calling `settingsChanged` via the `onDidChangeConfiguration` hook, the fact that `preferredModel` will change before models are refreshed presents an issue - it'll never be able to detect that change.

dividedmind (approved):

PR #1075, opened 2025-03-14
fix: Propagate Copilot content filter errors

dustinbyrne:
This change handles `Error: Response got filtered.`, propagating it to the OpenAI API client via the `content_filter` finish reason. This allows the client to notify the client that their response was terminated.

### DIFF
--- a/src/services/chatCompletion.ts
+++ b/src/services/chatCompletion.ts
@@ -413,8 +413,8 @@ async function streamChatCompletion(
   result: LanguageModelChatResponse,
   countTokens: () => Promise<number>
 ) {
+  const chunk = prepareChatCompletionChunk(model);
   try {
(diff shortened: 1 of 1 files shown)


src/services/chatCompletion.ts
```diff
@@ -427,7 +427,15 @@ async function streamChatCompletion(
     res.end();
   } catch (e) {
     warn(`Error streaming response: ${e}`);
-    if (isNativeError(e)) warn(e.stack);
+    if (isNativeError(e)) {
+      warn(e.stack);
+      if (e.message === 'Response got filtered.') {
+        chunk.choices[0].finish_reason = 'content_filter';
+        res.end(`data: ${JSON.stringify(chunk)}\n\n`);
```
dividedmind, line 434: Shouldn't it also write `data: [DONE]`?

PR #1106, opened 2025-11-23
feat: GitHub Releases for CLI & improved asset updates

### LINKED ISSUE #1105: VS Code extension uses outdated .appmap/bin/appmap binary after package update
**Problem Statement:**
A user encountered an issue where, after having an older AppMap binary version (3.190.0) in `.appmap/bin/appmap`, installing a new package with a more recent bundled version (3.196.0) did not update which binary VS Code uses. The extension continues to use the old binary, causing version mismatches and possible unexpected behavior.

**Steps to Reproduce:**
1. Copy an older AppMap binary (e.g., 3.190.0) to `.appmap/bin/appmap`.
2. Install (or upgrade) the package to one that bundles a newer version (e.g., 3.196.0).
3. Observe that the VS Code extension still references and uses the older binary in `.appmap/bin/appmap` instead of the new bundled version.

**Expected Behavior:**
VS Code extension should detect the presence of a newer bundled AppMap binary and use it in preference to any older binaries present in `.appmap/bin/appmap`.

**Actual Behavior:**
The extension defaults to the previous binary found in `.appmap/bin/appmap`, ignoring the more recent bundled version.

**Suggested Fix:**
- Review how the extension detects which AppMap binary to use.
- Prioritize using the most recent binary available—either bundled or in the workspace—based on the version number.
- Notify the user if an older version is being used when a newer one is available.

**Environment:**
- VS Code extension: getappmap/vscode-appland
- Original binary: 3.190.0 in `.appmap/bin/appmap`
- Bundled binary: 3.196.0 (installed via package)

dividedmind:
This PR introduces several improvements related to CLI tool version management, asset handling, and testing infrastructure.

Key changes include:

*   **feat: Use GitHub Releases for CLI version checks:** Updates the CLI tool download service to fetch the latest versions from the GitHub releases API instead of the npm registry. This improves reliability, especially in environments where npm might be blocked. Includes `GitHubReleaseResolver` for version resolution, caching of GitHub API responses, and prioritization of GitHub releases for `appmap` and `scanner` CLI tools. Tests have been updated to mock the GitHub releases API and handle caching.
*   **fix(assets): Unconditionally update asset symlinks and copies:** The `AssetService.ensureLinks` method now unconditionally updates symlinks and copies to ensure the most recent asset version is always used, even if an older version already exists. This resolves an issue where older assets were not being replaced by newer bundled or cached versions. A test case has also been added. (Fixes #1105)
*   **test: Consolidate BundledFileDownloadUrlResolver mocking in tests:** Consolidate the mocking of
`BundledFileDownloadUrlResolver.extensionDirectory` to the `beforeEach` block in `assetService.test.ts`. This ensures test environment isolation and consistency, resolving issues where `JavaAgentDownloader` was not downloading the Java agent due to incorrect asset resolution.
*   **tooling: Add AGENTS.md file:** Adds an AGENTS.md file for agent-specific instructions.

### DIFF
--- a/src/assets/index.ts
+++ b/src/assets/index.ts
@@ -58,17 +58,51 @@ class StaticVersionResolver implements VersionResolver {
   }
 }
 
-class GitHubVersionResolver implements VersionResolver {
-  constructor(private readonly repo: string) {}
+interface GitHubRelease {
+  tag_name: string;
+}
 
-  async getLatestVersion(): Promise<string | undefined> {
-    const res = await tryRequest(`https://api.github.com/repos/${this.repo}/releases/latest`);
+// exported mainly for testing purposes
+export const GithubReleaseCache = new Map<string, Promise<GitHubRelease[] | undefined>>();
+
+async function fetchGitHubReleases(repo: string): Promise<GitHubRelease[] | undefined> {
+  if (GithubReleaseCache.has(repo)) {
+    return GithubReleaseCache.get(repo);
+  }
+
+  const promise = (async () => {
+    const res = await tryRequest(`https://api.github.com/repos/${repo}/releases`);
     if (res) {
       try {
-        const json: { tag_name?: string } = (await res.json()) as Record<string, unknown>;
-        return json.tag_name?.replace(/^v/, '');
-      } catch (e: unknown) {
-        log.warning(`Failed to retrieve ${this.repo} version from GitHub: ${e}`);
+        const json = (await res.json()) as GitHubRelease[];
+        if (Array.isArray(json)) {
+          return json.map((release) => ({ tag_name: release.tag_name }));
+        }
+      } catch (e) {
+        log.warning(`Failed to retrieve ${repo} releases from GitHub: ${e}`);
+      }
+    }
+    return undefined;
+  })();
+
+  GithubReleaseCache.set(repo, promise);
+  return promise;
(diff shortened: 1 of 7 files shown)


src/assets/index.ts
```diff
@@ -58,17 +58,56 @@ class StaticVersionResolver implements VersionResolver {
...
+  constructor(private readonly repo: string, private readonly prefix?: string) {}
+
+  async getLatestVersion(): Promise<string | undefined> {
+    const releases = await fetchGitHubReleases(this.repo);
+    if (!releases) return undefined;
+
+    for (const release of releases) {
+      if (!release.tag_name) continue;
+
+      if (this.prefix) {
```
hleb-rubanau: Just a remark -- if we may ever use prerelease versions for assets (which I honestly recommend to consider, because sooner or later any opensource product comes to separating 'most stable' vs 'most recent' timelines), the resolver should explicitly discard them, only considering strict match towards '\d+.\d+.\d+$'. 

It would be trivial but useful check -- so I recommend it without insisting.

hleb-rubanau (approved):
All makes sense, agentic instructions are nice (suggesting transformation to Claude 'skills' in the future, if considerable). 

The only remark I left is recommendation, which is not a blocker for the release.

PR #1107, opened 2026-01-15
feat: Fix a code issue from finding or known function

kgilpin:
Adds an inline markdown icon to finding rows that opens a comprehensive finding report as a   
markdown preview with clickable file links.

Intended usage:
- Paste into an AI Assistant to fix the problem.
                                                                                              
Key features:                                                                                 
- Opens finding synopsis as markdown preview with workspace-relative clickable links          
- Shows complete rule context (ID, description, documentation)                                
- Includes stack trace, SQL queries, and participating events without truncation              
- Automatic line number validation (strips invalid values)                                    
                                                                                              
Technical:                                                                                    
- Created RawEventData type definitions for proper typing of snake_case event data            
- Fixed type mismatches in Finding interface fields 

# Example

Screenshot 2026-01-15 at 2 41 55 PM

# N plus 1 SQL query

This AppMap matches the following analysis rule: `n-plus-one-query`

Finds occurrences of a query being repeated within a loop.

**Impact Domain**: Performance | **Occurrences**: 8

## Finding Details
```
<templates>/UsersKgilpinSourceLandOfAppsDjangoOscar_Issue2484SrcOscarTemplatesOscarCatalogueBrowseHtml#render[218]
contains 8 occurrences of SQL: SELECT "partner_stockrecord"."id",
"partner_stockrecord"."product_id", "partner_stockrecord"."partner_id",
"partner_stockrecord"."partner_sku", "partner_stockrecord"."price_currency",
"partner_stockrecord"."price", "partner_stockrecord"."num_in_stock",
"partner_stockrecord"."num_allocated", "partner_stockrecord"."low_stock_threshold",
"partner_stockrecord"."date_created", "partner_stockrecord"."date_updated" FROM
"partner_stockrecord" WHERE "partner_stockrecord"."product_id" = ? LIMIT ?
```

## Group Summary
```
SELECT "partner_stockrecord"."id", "partner_stockrecord"."product_id",
"partner_stockrecord"."partner_id", "partner_stockrecord"."partner_sku",
"partner_stockrecord"."price_currency", "partner_stockrecord"."price",
"partner_stockrecord"."num_in_stock", "partner_stockrecord"."num_allocated",
"partner_stockrecord"."low_stock_threshold", "partner_stockrecord"."date_created",
"partner_stockrecord"."date_updated" FROM "partner_stockrecord" WHERE
"partner_stockrecord"."product_id" = ? LIMIT ?
```

## Source Location
[src/oscar/apps/partner/strategy.py:199](src/oscar/apps/partner/strategy.py#L199)

## Stack Trace
- [src/oscar/apps/partner/strategy.py:200](src/oscar/apps/partner/strategy.py#L200)
- [src/oscar/apps/partner/strategy.py:152](src/oscar/apps/partner/strategy.py#L152)
- [src/oscar/apps/partner/strategy.py:135](src/oscar/apps/partner/strategy.py#L135)

## Code Event Context
- **commonAncestor**: <templates>.UsersKgilpinSourceLandOfAppsDjangoOscar_Issue2484SrcOscarTemplatesOscarCatalogueBrowseHtml#render

## AppMap
[tmp/appmap/requests/1767987869_01673_http_127_0_0_1_8000_en-gb_catalogue_.appmap.json](tmp/appmap/requests/1767987869_01673_http_127_0_0_1_8000_en-gb_catalogue_.appmap.json)

## References
- https://appland.com/docs/analysis/rules-reference.html#n-plus-one-query
- [CWE-1073](https://cwe.mitre.org/data/definitions/1073.html)

### DIFF
--- a/src/commands/openFindingAsMarkdown.ts
+++ b/src/commands/openFindingAsMarkdown.ts
@@ -1,0 +1,232 @@
+import * as vscode from 'vscode';
+import * as path from 'path';
+import { FindingTreeItem } from '../tree/findingsTreeDataProvider';
+import { RawEventData } from '../types/rawEvent';
+
+const MAX_STACK_FRAMES = 5;
+const MESSAGE_WIDTH = 100;
+
+function wordWrap(text: string, maxWidth: number): string {
+  const lines: string[] = [];
+  const paragraphs = text.split('\n');
+
+  for (const paragraph of paragraphs) {
+    if (paragraph.trim().length === 0) {
+      lines.push('');
+      continue;
+    }
+
+    const words = paragraph.split(/\s+/);
+    let currentLine = '';
+
+    for (const word of words) {
+      const testLine = currentLine.length === 0 ? word : `${currentLine} ${word}`;
+
+      if (testLine.length <= maxWidth) {
+        currentLine = testLine;
+      } else {
+        if (currentLine.length > 0) {
+          lines.push(currentLine);
+        }
+        currentLine = word;
+      }
+    }
+
+    if (currentLine.length > 0) {
+      lines.push(currentLine);
+    }
+  }
+
+  return lines.join('\n');
+}
+
+function createFileLink(stackFrame: string, workspaceFolder: vscode.WorkspaceFolder): string {
+  // Parse format: "path/to/file.ts:42" or "path/to/file.ts" or "path/to/file.ts:-1"
+  const match = stackFrame.match(/^(.+?)(?::(-?\d+))?$/);
+  if (!match) return stackFrame;
+
+  const [, filePath, lineStr] = match;
+  let lineNumber: number | null = null;
+
+  // Parse and validate line number
+  if (lineStr) {
+    const parsedLine = parseInt(lineStr, 10);
+    // Only use line number if it's a positive integer (>= 1)
+    if (parsedLine >= 1) {
+      lineNumber = parsedLine;
+    }
+  }
+
+  // Resolve to absolute path first
+  const absolutePath = path.isAbsolute(filePath)
+    ? filePath
+    : path.join(workspaceFolder.uri.fsPath, filePath);
+
+  // Convert to workspace-relative path for both display and link
+  const relativePath = vscode.workspace.asRelativePath(absolutePath);
+
+  // Create link URL (workspace-relative with line number fragment if present)
+  const linkUrl = lineNumber ? `${relativePath}#L${lineNumber}` : relativePath;
+
+  // Create display text (with line number after colon if present)
+  const displayText = lineNumber ? `${relativePath}:${lineNumber}` : relativePath;
+
+  return `[${displayText}](${linkUrl})`;
+}
+
+function formatFindingSynopsis(
+  findingTreeItem: FindingTreeItem,
+  workspaceFolder: vscode.WorkspaceFolder
+): string {
+  const resolvedFinding = findingTreeItem.finding;
+  const finding = resolvedFinding.finding;
+
+  // Build synopsis in markdown format
+  const parts: string[] = [];
+
+  // Header with rule title
+  parts.push(`# ${finding.ruleTitle}`);
+  parts.push('');
+
+  // Rule ID
+  parts.push(`This AppMap matches the following analysis rule: \`${finding.ruleId}\``);
+  parts.push('');
+
+  // Rule description
+  if (resolvedFinding.rule.description) {
+    parts.push(resolvedFinding.rule.description);
+    parts.push('');
+  }
+
+  // Impact domain and occurrence count
+  const metadata: string[] = [];
+  if (resolvedFinding.impactDomain) {
+    metadata.push(`**Impact Domain**: ${resolvedFinding.impactDomain}`);
+  }
+  if (finding.occurranceCount && finding.occurranceCount > 1) {
+    metadata.push(`**Occurrences**: ${finding.occurranceCount}`);
+  }
+  if (metadata.length > 0) {
+    parts.push(metadata.join(' | '));
+    parts.push('');
+  }
+
+  // Message (word-wrapped in code block)
+  parts.push(`## Finding Details`);
+  parts.push('```');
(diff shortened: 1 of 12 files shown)


src/lib/promptForAppMap.ts
```diff
@@ -43,17 +43,20 @@ export async function promptForAppMap(
...
+          return item;
         })
     )
   ).sort((a, b) => a.label.localeCompare(b.label));
-  const result = await vscode.window.showQuickPick<AppMapQuickPickItem>(items);
+  const result = await vscode.window.showQuickPick(items);
   if (!result) return;
 
-  return result.resourceUri;
+  return uriMap.get(result.detail!);
```
dividedmind, line 61: Note we could instead just rename the property on the quick pick item so that wouldn't clash with the resourceUri proposal, eg. `appmapResourceUri` — showQuickPick accepts anything that extends a QuickPickItem and would just ignore unknown properties. (I find it mildly surprising that including it caused it to fail anyway — it seems it's just used to provide default label, description and icon if not explicitly set?)

src/commands/openFindingAsMarkdown.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+  }
+
+  // Resolve to absolute path first
+  const absolutePath = path.isAbsolute(filePath)
+    ? filePath
+    : path.join(workspaceFolder.uri.fsPath, filePath);
+
+  // Convert to workspace-relative path for both display and link
+  const relativePath = vscode.workspace.asRelativePath(absolutePath);
```
dividedmind, line 66: This will include workspace folder name if there are multiple folders. It's great for display, but I wonder: won't this break the link? I'm not sure how VSCode handles markdown links in this case.

In addition, what happens if the file is absolute and somehow in a different workspace folder than provided in the argument? Should this function validate it?

src/commands/openFindingAsMarkdown.ts
```diff
@@ -0,0 +1,232 @@
...
+  parts.push(`## References`);
+  if (resolvedFinding.rule.references && Object.keys(resolvedFinding.rule.references).length > 0) {
+    parts.push(`- ${resolvedFinding.rule.url}`);
+    for (const [name, url] of Object.entries(resolvedFinding.rule.references)) {
+      parts.push(`- [${name}](${url.toString()})`);
+    }
+    parts.push('');
+  }
+
+  return parts.join('\n');
```
dividedmind, line 200: Perhaps surprisingly, in v8 appending to a string is faster than pushing parts and then concatenating — strings are optimized for this case and this avoids the overhead of creating many objects. It might also be more readable.

src/commands/openFindingAsMarkdown.ts
```diff
@@ -0,0 +1,232 @@
...
+          .filter(Boolean)
+          .join('');
+        parts.push(`- **${name}**: ${location}`);
+      }
+    }
+    parts.push('');
+  }
+
+  // AppMap file (workspace-relative path as clickable link)
+  const workspaceRelativePath = vscode.workspace.asRelativePath(resolvedFinding.appMapUri.fsPath);
```
dividedmind, line 185: This potentially has the same issue as above when there are multiple folders.

src/commands/openFindingAsMarkdown.ts
```diff
@@ -0,0 +1,232 @@
...
+          const document = await vscode.workspace.openTextDocument({
+            content: synopsis,
+            language: 'markdown',
+          });
+
+          // Show the document first (required for preview command to work)
+          await vscode.window.showTextDocument(document);
+
+          // Then open the markdown preview
+          await vscode.commands.executeCommand('markdown.showPreview', document.uri);
```
dividedmind, line 223: Perhaps we don't need to await this one; IMO it's better if opening preview fails silently than showing a potentially confusing error message.

src/lib/promptForAppMap.ts
```diff
@@ -43,20 +43,17 @@ export async function promptForAppMap(
...
+          } as AppMapQuickPickItem;
         })
     )
   ).sort((a, b) => a.label.localeCompare(b.label));
-  const result = await vscode.window.showQuickPick(items);
+  const result = await vscode.window.showQuickPick<AppMapQuickPickItem>(items);
   if (!result) return;
 
-  return uriMap.get(result.detail!);
+  return result.resourceUri;
```
dividedmind: This just reverts the previous commit?

src/services/resolvedFinding.ts
```diff
@@ -187,13 +188,17 @@ export class ResolvedFinding {
...
     const state = {
       currentView: 'viewFlow',
-      selectedObject: `event:${finding.event.id}`,
+      selectedObject: `event:${rawEvent.id}`,
     } as any; // eslint-disable-line @typescript-eslint/no-explicit-any
 
     if (finding.relatedEvents) {
-      state.traceFilter = finding.relatedEvents.map((evt) => ['id', evt.id].join(':')).join(' ');
+      const rawRelatedEvents = finding.relatedEvents as unknown as RawEventData[];
+      state.traceFilter = rawRelatedEvents.map((evt) => ['id', evt.id].join(':')).join(' ');
```
dividedmind, line 201: The code in this function only uses `id` in events which should be fine either way. I don't understand the reason for this change.

src/types/rawEvent.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * RawEventData represents the actual JSON structure of events as stored in AppMap files
+ * and returned from the scanner. These use snake_case naming (from Ruby/Python origins),
+ * unlike the Event class from @appland/models which uses camelCase.
+ *
+ * Note: The Finding interface from @appland/scanner incorrectly types some fields as Event
+ * when they are actually raw JSON objects. This type provides the correct structure.
+ */
```
dividedmind, line 8: If we ever fix it there to actually return what is advertised in the typing then this code will break (BTW, we should probably open an issue or PR there). It would be more robust to have a wrapper that fixes up the objects instead.

dividedmind (changes requested):

PR #1108, opened 2026-04-18
Fixes for model configuration

dividedmind:
This fixes the issue where updating OpenAI API base URL through the 'gear' icon in UI was impossible — the changes were simply ignored. Looking into this exposed two bugs:
- the model config change webview event handler for some reason only dispatched configuration update when a _secret_ setting (such as an API key) was updated;
- fixing the above revealed that deleting environment variables (such as when the user would like to go back to the default OpenAI endpoint) didn't work due to how configuration objects returned by VSCode behave.

To manually test this change, simply open the gear icon and try setting OpenAI base URL; observe that it's updated in the environment variables setting in configuration and that AppMap RPC is restarted. Then, open the gear icon again, clear the override; observe that the variable is gone from the setting and the RPC server is restarted again.

### DIFF
--- a/src/services/rpcProcessService.ts
+++ b/src/services/rpcProcessService.ts
@@ -295,7 +295,11 @@ export default class RpcProcessService implements Disposable {
     }
 
     if (change.env) {
-      const env = vscode.workspace.getConfiguration('appMap').get('commandLineEnvironment', {});
+      // we need to clone the returned object because it's
+      // a proxy and we can't delete properties from it directly
+      const env = {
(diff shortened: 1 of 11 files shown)


kgilpin (approved):
Nice fix. That is a lot of test refactoring!

dividedmind:
> Nice fix. That is a lot of test refactoring!

Thanks! If there's one single uncontroversial benefit of using AIs for development, it's how easy they make it to keep test suites in shape :)

PR #1125, opened 2026-09-11
feat: install and update AppMap agent skills

kgilpin:
## Summary

Installs the AppMap agent skills from the latest `getappmap/skills` GitHub release for Claude Code and GitHub Copilot, keeps them up to date, and offers the AppMap MCP servers to each open workspace.

**How it works**

```
appMap.skills.install = true | false        (default: true, machine-scoped)

  true   -> SkillsCache   unpack latest release at ~/.appmap/skills/<skill> (+ .version)
            SkillLink     link each skill into every appMap.skills.directories entry
                          default: ~/.claude/skills, ~/.agents/skills
            first time skills land on disk: one notification, [OK] [Uninstall]
  false  -> nothing

independently, per open workspace:
  .vscode/mcp.json lacks "appmap" or "appmap-gold-traces"
         -> offer: [Add] [Not now] [Don't ask again]
```

**Skills**

- `~/.claude/skills` is read by Claude Code and by Copilot in VS Code. `~/.agents/skills` is read by every Copilot surface. Together they cover both agents with no per-repository installs. An empty directory list keeps the cache current without linking anywhere, for people who place the skills by hand.
- **Never overwrites user content.** A skill entry is only touched if it is absent, a symlink into our cache, or a copy carrying our marker file. Links to skills dropped from a release are removed. Anything else is left alone.
- A `.appmap-skills` note in each directory names the entries we manage, where they come from, and the setting that turns them off.
- The user is told once per machine when skills first land on disk. Uninstall is confirmed separately, spells out what is lost, removes our links, and turns the setting off.
- Falls back to a copy where symlinks are unavailable (Windows without Developer Mode). The copy is assembled beside its destination and renamed into place, so a failed copy leaves nothing behind.
- The three skills settings are machine-scoped, so a workspace's `.vscode/settings.json` cannot redirect what gets linked into the user's agent directories.
- The install runs in the background and never blocks Navie. It is skipped under integration tests, which run against the real home directory.
- `runUpdates` now takes a lock path. Tools and skills lock separately, since a caller that finds the lock held assumes the holder is doing its work and skips.

**MCP servers**

- Two entries are offered: `appmap` (`appmap query mcp`) and `appmap-gold-traces` (the same, pointed at `gold_traces/baseline/appmaps` with its own query database). Both use `${userHome}/.appmap/bin/appmap`, which VS Code expands, so the file is portable and does not depend on PATH.
- The file is in the user's repository, so it is never written without asking. Only missing entries are offered and added; an existing entry, however configured, is never changed. The file is edited in place so other servers, comments, and formatting survive.
- The offer does not depend on the skills setting, since Copilot in VS Code reads `mcp.json` without any skills. "Don't ask again" is remembered per workspace in workspace state. A failure to add is shown to the user.

**Structure** (`src/services/`)

- `skills/skillsCache.ts`: download, extraction, cache layout.
- `skills/skillLink.ts`: what an installed skill looks like on disk, reported as `absent | foreign | link | copy`, the sync loop, removal, and the note file.
- `skills/mcpConfig.ts`: which AppMap servers a workspace's `mcp.json` is missing, and adding them.
- `skillService.ts`: settings, notifications, orchestration.

## Notification

Screenshot 2026-09-18 at 3 01 42 PM

## Notes

- Skill downloads are not digest-verified; they rely on TLS and GitHub. See `src/assets/FOLLOWUPS.md` item 5.
- Adds `tar` and `jsonc-parser` as runtime dependencies.
- Fixes a race in `CommandRegistry.commandReady` that made the install-agent integration test hang once the skills install added concurrent work at activation.

## Test plan

- [x] `yarn test:unit` passes (328 tests). `test/unit/services/skillService.test.ts` and `skills/mcpConfig.test.ts` cover install into multiple directories, an empty directory list, upgrade in place, removal of retired skills, user-owned entries left alone, one bad directory not blocking the others, copy fallback, refresh, and all-or-nothing copy, lock isolation from the tools updater, the note file, the first-install notification and uninstall flow, the MCP offer (add both, add only the missing one, dismiss, decline for good, already present, unparsable file reported), and the integration-test skip.
- [x] CI green, including the integration suite.
- [ ] Fresh activation installs into `~/.claude/skills` and `~/.agents/skills` and shows the one-time notification
- [ ] Uninstall from that notification confirms, removes the links, and sets `appMap.skills.install` to false
- [x] Opening a workspace without the MCP entries adds them and shows a notification
- [x] A workspace with a hand-written `appmap` entry is offered only `appmap-gold-traces`, and the existing entry is untouched
- [ ] Windows without Developer Mode falls back to copy and a later release replaces the copy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0111ViDx3qj7nssryzc9ZtvZ


src/services/skillService.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  return false;
+}
+
+async function currentInstalledVersion(linkPath: string): Promise<string | undefined> {
+  try {
+    const stats = await lstat(linkPath);
+    if (stats.isSymbolicLink()) {
+      const target = resolvePath(await readlink(linkPath));
+      const relative = target.slice(AppMapSkillsDir().length + 1);
```
dividedmind: This assumes the target is managed; in the current flow it's always true, but it's easy to overlook in case things shift. Perhaps it would be more robust to have a single function do both – eg. return managed version or undefined if not managed? Looking at the call sites it should be a simple replacement. As a bonus, we won't have to resolve the link twice. The only corner case is if the directory doesn't exist yet, but perhaps it could be handled separately, semantically 'absent' and 'present and ours' seem like a different case.

src/services/skillService.ts
```diff
@@ -0,0 +1,234 @@
...
+  // Extract into a scratch directory first so a failed or partial extraction
+  // never leaves a half-populated version directory that looks complete.
+  const stagingDir = `${versionDir}.staging`;
+  await rm(stagingDir, { recursive: true, force: true });
+  await mkdir(stagingDir, { recursive: true });
+  try {
+    // strip: 1 drops the `skills-<version>/` prefix GitHub adds to source archives.
+    await tar.extract({ file: archivePath, cwd: stagingDir, strip: 1 });
+    await rm(versionDir, { recursive: true, force: true });
+    await cp(stagingDir, versionDir, { recursive: true });
```
dividedmind: Why not mv?

src/services/skillService.ts
```diff
@@ -0,0 +1,234 @@
...
+    );
+  }
+
+  static async ensureInstalled(throwOnError = false): Promise<void> {
+    if (!ExtensionSettings.autoUpdateSkills) {
+      log.info('Automatic AppMap skills updates are disabled, skipping.');
+      return;
+    }
+
+    return runUpdates([() => this.installLatest()], throwOnError);
```
dividedmind: `runUpdates` uses the same lock file for both the tool and the skill updates. What happens if they try to run concurrently? IIUC, the current lockFileSynchronizer implementation assumes there's only a single update process that's under the lock and if the lock is held by someone else, the update is never performed and instead the process just waits for the lock to be released without additional checking. If I'm right, this means if the skill update is fired when the lock is held by the tool update, the skill update won't be performed.

package.json
```diff
@@ -397,6 +389,16 @@
           "default": true,
           "description": "Enable automatic updates and downloads of AppMap tools."
         },
+        "appMap.autoUpdateSkills": {
```
dividedmind: Consider `appMap.skills.updateAutomatically` or something similar so both options are grouped together. Note that this is at odds with consistency with autoUpdateTools – not sure which is better.

src/services/skillService.ts
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+// Point `linkPath` at `skillDir`. Prefers a directory symlink; falls back to a
+// recursive copy on platforms that disallow symlink creation (Windows without
+// Developer Mode), stamping the copy with a marker so we can recognize and
+// update it later.
+async function installSkill(skillDir: string, linkPath: string, version: string): Promise<void> {
+  await rm(linkPath, { recursive: true, force: true });
+  await mkdir(ClaudeSkillsDir(), { recursive: true });
```
dividedmind: Why is ClaudeSkillsDir created here? `installSkill` can take a different linkPath. Either create the skills dir outside of this function, or mkdir the linkPath parent (or take the parent and name, or just the parent and take the name from the source); otherwise it's mixing concerns.

dividedmind:
Functionally I think it's alright, although I'd be wary of activating it by default without asking the user first. I'd be surprised by an extension messing with my claude dir without explicit permission. In fact, I think keeping skills up to date and linking them into ~/.claude should be a separate concern; I might want the former (could be on by default) while preferring to do symlinking myself. In the current design this would be hindered by the lack of stable path – `~/.appmap/skills/<version>` doesn't lend itself to manual management. I think it would work better to just keep skills at ~/.appmap/skills without the version – if we're not keeping the archived versions around anyway the extra directory layer doesn't buy anything, just makes manual management like this more difficult. Just unpack in ~/.appmap/skills.staging, if unpack succeeded, rm && mv, add a marker file with the version number to prevent redownloading.

kgilpin:
I don't see why adding new skills to your machine would be something you'd consider intrusive or needs approval? We just need the auto-installer to detect if you've somehow taken over management of the appmap skills yourself.

kgilpin:
> Functionally I think it's alright, although I'd be wary of activating it by default without asking the user first. I'd be surprised by an extension messing with my claude dir without explicit permission. In fact, I think keeping skills up to date and linking them into ~/.claude should be a separate concern; I might want the former (could be on by default) while preferring to do symlinking myself. In the current design this would be hindered by the lack of stable path – `~/.appmap/skills/<version>` doesn't lend itself to manual management. I think it would work better to just keep skills at ~/.appmap/skills without the version – if we're not keeping the archived versions around anyway the extra directory layer doesn't buy anything, just makes manual management like this more difficult. Just unpack in ~/.appmap/skills.staging, if unpack succeeded, rm && mv, add a marker file with the version number to prevent redownloading.

If you change the skill links from the managed symlink, it leaves your manual change alone. So IMO we are already good here. As soon as you take over management of the skills, the extension stops managing them.

package.json
kgilpin: Addressed

src/services/skillService.ts
kgilpin: Addressed

src/services/skillService.ts
kgilpin: The swap in SkillsCache.update at src/services/skills/skillsCache.ts:67 is now two renames:
```
await rm(previous, { recursive: true, force: true });
if (await exists(this.dir)) await rename(this.dir, previous);
await rename(staging, this.dir);
await rm(previous, { recursive: true, force: true });
```

The staging directory is a sibling of the cache, ~/.appmap/skills.staging, so it is always on the same filesystem and the rename is a metadata operation. The old contents move aside to ~/.appmap/skills.previous rather than being deleted first, so there is no moment where the cache has been emptied but not yet refilled. Before the swap, the staged release is checked to contain at least one skill, and the version marker is written into staging, so whatever lands at the final path is complete.

src/services/skillService.ts
kgilpin: The two functions collapsed into one inspect() method on SkillLink, at src/services/skills/skillLink.ts:28. It does one lstat and at most one readlink, and returns one of four states.

dividedmind:
> I don't see why adding new skills to your machine would be something you'd consider intrusive or needs approval? We just need the auto-installer to detect if you've somehow taken over management of the appmap skills yourself.

I think we should avoid changing the user global configuration of unrelated software without approval – this can be surprising in a bad way. But I agree we need to make it easier and more frictionless for the users to install the skills.

I propose:
1. Always keeping the skills in ~/.appmap/skills updated, on startup (but maybe not surface error messages for these automatic updates without explicit user action, we don't want offline users that otherwise have CLI already installed to have a bad experience) – this is ours and the user might want to point the agent there even without installing the scripts (you can just tell an agent about a skill location explicitly and it will read and use it).
2. Surfacing this to the user – whenever the automatic install/update succeeds, show a notification, something like "AppMap supplies skills that make it easier for AI agents to work with it. They're automatically installed and kept up to date in ~/.appmap/skills. Would you like to install them in your ~/.claude/skills and ~/.agents/skills to be globally available? [Yes] [No] [Read more] [Never ask again]" – 'yes' answer should remember to autoupdate, 'no' should ask again, 'never' should disable this notification; 'read more' should link to some docs if possible. Nothing clicked – same as 'no' – will ask on the next startup.
   - Alternatively (less preferable IMO) install by default (although this complicates integration tests – we need to take care to skip this in integration test context) and show a notification like "AppMap supplies skills that make it easier for AI agents to work with it. They're automatically installed and kept up to date in ~/.appmap/skills. We've also installed them in ~/.claude/skills and ~/.agents/skills so they're globally available. [Ok] [Read more] [Revert]" – 'ok' keeps them up to date and never mentions it again, unless the user removes the links/copies themselves; 'read more' is again a link to the docs; 'revert' removes the links and doesn't mention it again. Nothing clicked – will notify on the next startup.
3. An explicit command to install the skills, "install appmap skills"; should ask the user to confirm the target directories. This should probably be invoked under the hood by the above (with or without target directory confirmation in this case? not sure). It's fine to autoupdate later, unless the user removes the links/copies themselves.

WDYT?

dividedmind:
Noting for later, not blocking: the `skills.directories` default lists `~/.claude/skills` and `~/.agents/skills`, but Claude Code only reads `~/.claude/skills` for personal skills, and everything that does read `~/.agents/skills` (Copilot in VS Code) reads `~/.claude/skills` too — so the second entry adds no coverage. Harmless since Copilot dedupes, just redundant.

The case it doesn't cover: a Copilot-only user who has never run a CLI agent. We create `~/.claude` for someone who doesn't use Claude Code. `~/.copilot/skills` is Copilot's own directory and VS Code reads it as well, so it'd be the friendlier target there. Worth a look next time we touch the defaults.

Refs: [VS Code](https://code.visualstudio.com/docs/agent-customization/agent-skills) reads `~/.copilot/skills`, `~/.claude/skills` and `~/.agents/skills`; [Claude Code](https://code.claude.com/docs/en/agent-sdk/skills) reads only `~/.claude/skills`.

src/services/skills/mcpConfig.ts
```diff
@@ -0,0 +1,65 @@
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
+import { applyEdits, modify, parse, ParseError } from 'jsonc-parser';
+
+// The entry we want present under "servers" in .vscode/mcp.json.
+export const APPMAP_MCP_SERVER = {
+  type: 'stdio',
+  command: 'appmap',
```
dividedmind: `.vscode/mcp.json` gets `"command": "appmap"`, resolved from `PATH`. But the extension installs the CLI at `~/.appmap/bin/appmap` and nothing adds that directory to `PATH` — every other call site in the extension invokes the CLI by absolute path. So for anyone whose only AppMap CLI came from the extension, the entry we just talked them into committing won't start, and the failure lands on their teammates too since the file is checked in.

No obvious right answer, which is why I'm not fixing it here: an absolute path can't go in a committed file, checking `PATH` before offering silently disables the feature for most users, and putting `~/.appmap/bin` on `PATH` is a bigger change than this PR. Flagging for a decision.

dividedmind (approved):
I added a bunch of bugfixes, a provenance note about the skills and a notification for the user on first install. Feel free to merge, but you might consider removing the mcp part – it currently just adds `appmap` command which will be dead for users that don't have it in PATH (presumably most).

kgilpin:
It can be done like this:

"command": "${userHome}/.appmap/bin/appmap",

src/services/skills/mcpConfig.ts
kgilpin: It now writes: 

```
"command": "${userHome}/.appmap/bin/appmap"
```

